PR #1, opened 2018-12-16
Read configuration files into a single dictionary

susam:
This change adds support for reading multiple configuration files in
YAML format into a single dictionary. The configuration files can be
specified using command line options (`-c` or `--config` option).

When no configuration file option is specified in the command line, the
framework looks for `config.base.yaml` and `config.yaml` in the current
directory. When multiple configuration files are specified, the
configuration in the file that is specified later is merged with the
configuration in the files specified earlier.


README.md
```diff
@@ -2,3 +2,35 @@ Cloud Marker
...
+
+ 1. Ensure Python 3 is installed.
+
+        # On macOS
+        brew install python
+
+ 2. Clone the project repository.
+
+        git clone https://github.com/cloudmarker/cloudmarker.git
+
```
s-nirali: You might want to add the make command `make venv` and `. ./venv` here for setting up the project.

pylama.ini
```diff
@@ -0,0 +1,16 @@
+[pylama]
+linters = pycodestyle,pydocstyle,pyflakes,pylint
+ignore = D203,D213,D406,D407,C0103,W0511
```
s-nirali: I think that the pip install command automatically adds the following linters as well, mccabe and isort. You might want to add these here.

config.base.yaml
```diff
@@ -0,0 +1,18 @@
...
+clouds:
+  mockcloud:
+    plugin: cloudmarker.clouds.mockcloud.MockCloud
+
+mockjob:
+  clouds:
+    - mockcloud
+  stores:
+    - filestore
+
```
s-nirali: In this structure, is the stores config like a default? Would the clouds also in future support specifying their own specific stores?

PR #3, opened 2018-12-18
Add load_plugin() function in utils module

susam:
The load_plugin() function will be used to instantiate plugin classes
with parameters specified in the configuration.

### DIFF
--- a/cloudmarker/util.py
+++ b/cloudmarker/util.py
@@ -3,6 +3,7 @@
 
 import argparse
 import copy
+import importlib
 import os
 
 import yaml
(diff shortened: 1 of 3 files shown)


cloudmarker/util.py
```diff
@@ -32,6 +33,51 @@ def load_config(config_paths):
...
+
+    The plugin_config parameter must be a dictionary with the following
+    keys:
+
+      - 'plugin': The value for this key must be a string that
+        represents the fully qualified class name of the plugin. The
+        fully qualified class name is in the dotted notation, e.g.,
+        ``pkg.module.ClassName``.
+      - 'params': The value for this key must be a dict that represents
+        the parameters to be passed to the ``__init__`` functio of the
```
s-nirali: `function` typo

cloudmarker/util.py
```diff
@@ -32,6 +33,51 @@ def load_config(config_paths):
...
+    # parts: module name and class name.
+    if len(parts) < 2:
+        msg = ('Invalid plugin class name: {}; expected format: '
+               '[<pkg>.]<module>.<class>'.format(plugin_config['plugin']))
+        raise PluginError(msg)
+
+    # Load the specified adapter class from the specified module.
+    plugin_module = importlib.import_module(parts[0])
+    plugin_class = getattr(plugin_module, parts[1])
+
```
s-nirali: I think this part of code should raise an error if the class name looks valid but does not exist. There could be a corresponding test case for it.

cloudmarker/util.py
susam: Updated the commit and force-pushed it. Thanks for noticing this.

cloudmarker/util.py
susam: Yes, it would lead to `AttributeError` because the attribute (the class we are looking for) would be missing from the module. I decided not to test this earlier because it is akin to testing whether the Python language is working properly or not.

This is one of those gray areas where I find it difficult to decide whether *to test or not to test*.

Now that you raise this point, I have now decided to add a test for it with the intention of *documenting the expected behavior as a test* in case of a missing plugin.

PR #7, opened 2018-12-27
Ignore proprietary format coverage files created

prateeknischal:
when running the command
. ./venv && coverage run -m unittest -v
creates temp files of the format .coverage.HOSTNAME.NUM.NUM

Modified the Makefile to use a ".coverage.*" pattern to remove files and
updated the same in .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Ignore automatically generated files.
 venv
 .coverage
+.coverage.*
 htmlcov/
 
 # Ignore local development files.
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,6 @@ checks: test coverage lint
 clean: FORCE
 	find . -name "__pycache__" -exec rm -r {} +
 	find . -name "*.pyc" -exec rm {} +
-	rm -rf .coverage htmlcov
+	rm -rf .coverage.* .coverage htmlcov
 
 FORCE:


prateeknischal:
I see that the coverage combine does remove all the junk files. It is up to the reviewer to reject this pull request.

.gitignore
```diff
@@ -1,6 +1,6 @@
 # Ignore automatically generated files.
 venv
-.coverage
+.coverage*
```
susam: How can I reproduce the creation of these files?

I was not aware that apart from `.coverage`, there are more `.coverage*` files being created. I would like to reproduce this once.

Makefile
```diff
@@ -52,6 +52,6 @@ checks: test coverage lint
 clean: FORCE
 	find . -name "__pycache__" -exec rm -r {} +
 	find . -name "*.pyc" -exec rm {} +
-	rm -rf .coverage htmlcov
+	rm -rf .coverage.* htmlcov
```
susam: I think the same glob pattern should be used in both `.gitignore` and `Makefile` for the sake of consistency. Right now, the two glob patterns differ by a `.`.

If you agree, could you please update this pull request?

.gitignore
susam: > How can I reproduce the creation of these files?
> 
> I was not aware that apart from `.coverage`, there are more `.coverage*` files being created. I would like to reproduce this once.

Never mind. I found the answer to my question in your commit message. Thanks for writing the commit message in a detailed manner. I could indeed reproduce the creation of `.coverage.HOSTNAME.NUM.NUM` files.

So I am going to suggest that we update the glob pattern in `.gitignore` to `.coverage.*` so that it matches the one in `Makefile`.

Issue #15, opened 2019-01-08
FirewallRule Anomaly Checker


r0hi7:
Create a very minimal yet complete, rule comparison based FirewallRule Checker.

PR #19, opened 2019-01-17
Add Firewall Rule Check plugin

r0hi7:
This PR implements #15

### DIFF
--- a/cloudmarker/checks/firewallcheck.py
+++ b/cloudmarker/checks/firewallcheck.py
@@ -1,0 +1,117 @@
+"""Firewall Rule Check plugin for GCP Firewall rule record.
+
+This modules define a :class:`FirewallCheck` that identifies weak Firewall
+rules for Google Cloud Platform.
+"""
+
+
+class FirewallCheck:
+    """Firewall Rule Check plugin.
+
+    This class identifies weak Firewall rules in the record objects by
+    comparing the them with predefined weak Firewall rules
+    configurations.
+    """
+
+    def __init__(self):
+        """Create instance of :class:`FirewallCheck` plugin."""
+        # self._rules defines the set of weak/misconfigured
+        # firewall rules. Given any firewall rule, is compared with
+        # these predefined weak misconfigurations if the rule matches to
+        # any one predefined rule, it will be marked as weak and will be
+        # reported as anomaly. Adding IPProtocol field have been skipped
+        # to check for misconfiguration as GCP docs says that: The
+        # protocol type is required when creating a firewall rule. This
+        # value can either be one of the following well known protocol
+        # strings (tcp, udp, icmp, esp, ah, ipip, sctp), or the IP
+        # protocol number, implying it cannot be *.
+        self._rules = {
+            'sourceRanges': [
+                '0.0.0.0/0',
+                '*'
+            ],
+            'destinationRanges': [
+                '0.0.0.0/0',
+                '*'
+            ],
+        }
+
+    def _process_rule(self, record):
+        misconfigurations = []
+
+        for weak_rule_key in self._rules:
+            if weak_rule_key in record:
+                for record_item in record[weak_rule_key]:
+                    if (record_item in
+                            self._rules[weak_rule_key]):
+                        misconfigurations.append(weak_rule_key
+                                                 + ':'
+                                                 + record_item)
+        return {
+            'misconfigurations': misconfigurations,
+            'id': record['id']
+        }
+
+    def eval(self, record):
+        """Evaluate Firewall rules for a given record.
+
+        Evaluate the weak security misconfigurations for each record and
+        collate each weakly identified record with its misconfigurations. This
+        function will only evaluate firewall_rule record.
+
+        For each such record it will create a dict object containing the weak
+        misconfigurations and return those records.
+
+        Arguments:
+            record (dict): Dictionary containing the rules definition.
+
+        Yields:
+            anomaly (dict): Dictionary representing anomaly.
+
+        The anomaly record yielded by this method would look like:
+
+        .. code:: python
+
+            {
+                "record_type": "firewall_alert",
+                "rule": "https://www.googleapis.com/compute/v1/projects"
+                        "/foo/global/firewalls/default-allow-http",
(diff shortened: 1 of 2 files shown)


cloudmarker/checks/firewallrulecheck.py
```diff
@@ -0,0 +1,119 @@
...
+        Yields:
+            anomaly (dict): Dictionary representing anomaly.
+
+        The anomaly record yielded by this method would look like:
+
+        .. code:: python
+
+            {
+                "record_type": "firewall_rule_anomaly",
+                "id": ""
```
susam: Please remove this line for `"id"` because I see there is a line for `"id"` a few lines below (line 86).

cloudmarker/checks/firewallrulecheck.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        # Process firewall rule only if the record type is 'firewall_rule' and
+        # the rule is not disabled.
+        if (record['record_type'] != 'firewall_rule'
+                or record['disabled']):
+            return
+
+        _anomaly_record = {
+            "record_type": "firewall_rule_anomaly",
+            "anomaly_type": "weak_firewall_rules",
```
susam: Please use single-quotes instead of double-quotes for single-line strings within Python code. This is the convention we are following in this project. Here is the PEP 8 recommendation on this: https://www.python.org/dev/peps/pep-0008/#string-quotes.

cloudmarker/checks/firewallrulecheck.py
```diff
@@ -0,0 +1,119 @@
...
+        # firewall rules. Given any firewall rule, is compared with
+        # these predefined weak misconfigurations if the rule matches to
+        # any one predefined rule, it will be marked as weak and will be
+        # reported as anomaly. Adding IPProtocol field have been skipped
+        # to check for misconfiguration as GCP docs says that: The
+        # protocol type is required when creating a firewall rule. This
+        # value can either be one of the following well known protocol
+        # strings (tcp, udp, icmp, esp, ah, ipip, sctp), or the IP
+        # protocol number, implying it cannot be *.
+        self._misconfigurations = {
```
susam: Please rename this variable to `self._rules` because this variable is defining the rules to detect misconfigurations.

cloudmarker/checks/firewallrulecheck.py
```diff
@@ -0,0 +1,119 @@
...
+                '*'
+            ],
+            'destinationRanges': [
+                '0.0.0.0/0',
+                '*'
+            ],
+        }
+
+    def _process_rule(self, record):
+        _misconfigurations = []
```
susam: This should be named as just `misconfigurations`. The underscore naming convention should not be used for local variables.

cloudmarker/checks/firewallrulecheck.py
```diff
@@ -0,0 +1,119 @@
...
+
+        for misconfig_field in self._misconfigurations:
+            if misconfig_field in record:
+                for record_misconfig_field in record[misconfig_field]:
+                    if (record_misconfig_field in
+                            self._misconfigurations[misconfig_field]):
+                        _misconfigurations.append(misconfig_field
+                                                  + ':'
+                                                  + record_misconfig_field)
```
susam: Please rename some of the variables in the above code so that it is easier to look at the variable and tell what it represents (a key or a value? within the rules dictionary or within the actual firewall record dictionary?). Here are my suggestions:

- `self._misconfigurations` to `self._rules`
- `misconfig_field` to `rule_key`
- `record_misconfig_field` to `record_item`

cloudmarker/checks/firewallrulecheck.py
```diff
@@ -0,0 +1,119 @@
...
+
+        """
+        # Process firewall rule only if the record type is 'firewall_rule' and
+        # the rule is not disabled.
+        if (record['record_type'] != 'firewall_rule'
+                or record['disabled']):
+            return
+
+        _anomaly_record = {
+            "record_type": "firewall_rule_anomaly",
```
susam: How about writing `firewall_rule_anomaly` as `firewall_alert`? The terminology I am thinking of following is this: *check modules* create *alert records*. The alert records are then sent to appropriate destinations by *alert modules*.

Similarly, `anomaly_type` could change to `alert_type`. But then again I am thinking if `alert_type` is really required. If it is not required, then perhaps we should remove it. We can always add it later if there is a good reason in future for why it is required.

cloudmarker/checks/firewallrulecheck.py
```diff
@@ -0,0 +1,119 @@
+"""Firewall Rule Check plugin for GCP Firewall rule record.
+
+This modules define a :class:`FirewallRuleCheck` that identifies weak Firewall
+rules for Google Cloud Platform.
+"""
+
```
susam: Please update the commit message to

```
Add FirewallCheck plugin
```
so that it matches the class name exactly and also that it is consistent with your previous commit message for GCPCloud plugin.

susam:
Please also add this to `config.base.json` under the `checks` key:

```
  firewallcheck:
    plugin: cloudmarker.checks.firewallcheck.FirewallCheck
```

This way we will have a ready-to-use check plugin named `firewallcheck` that can be used in `config.json`.

cloudmarker/checks/firewallrulecheck.py
r0hi7: How about `self._weak_rules`? as `_misconfigurations` defines weak rules.

cloudmarker/checks/firewallrulecheck.py
r0hi7: 👍

cloudmarker/checks/firewallrulecheck.py
susam: The rules in this dictionary are themselves are not weak. 😄 In fact, they are supposed to be *strong* rules that match or detect weak security configurations or security misconfigurations. Keeping this in mind, the names that seem appropriate are: `_detection_rules`, `_misconfiguration_detection_rules`, `_match_rules`, etc. You could choose a name along these lines or simply `_rules`.

PR #20, opened 2019-01-17
Add MongoDB Store plugin

r0hi7:


cloudmarker/stores/mongodbstore.py
```diff
@@ -0,0 +1,100 @@
...
+        self._mongodb_host = host
+        self._mongodb_port = port
+        self._mongodb_username = username
+        self._mongodb_password = password
+        self._db = db
+
+        self._buffer_size = buffer_size
+        self._buffer = {}
+
+        if not (username and password):
```
prateeknischal: `username` and `password` should be a required param as database will fail early instead of failing when the first `_flush()` will be called. 
```
>>> from pymongo import MongoClient
>>> client = MongoClient(host="0.0.0.0", connect=False)
>>> db = client["mongo-test"]
>>> db.list_collection_names()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/p0n002h/.venv/py-orm/lib/python3.7/site-packages/pymongo/database.py", line 690, in list_collection_names
    nameOnly=True)]
  File "/Users/p0n002h/.venv/py-orm/lib/python3.7/site-packages/pymongo/database.py", line 677, in list_collections
    **kwargs)
  File "/Users/p0n002h/.venv/py-orm/lib/python3.7/site-packages/pymongo/database.py", line 631, in _list_collections
    session=tmp_session)["cursor"]
  File "/Users/p0n002h/.venv/py-orm/lib/python3.7/site-packages/pymongo/database.py", line 514, in _command
    client=self.__client)
  File "/Users/p0n002h/.venv/py-orm/lib/python3.7/site-packages/pymongo/pool.py", line 579, in command
    unacknowledged=unacknowledged)
  File "/Users/p0n002h/.venv/py-orm/lib/python3.7/site-packages/pymongo/network.py", line 150, in command
    parse_write_concern_error=parse_write_concern_error)
  File "/Users/p0n002h/.venv/py-orm/lib/python3.7/site-packages/pymongo/helpers.py", line 155, in _check_command_response
    raise OperationFailure(msg % errmsg, code, response)
pymongo.errors.OperationFailure: command listCollections requires authentication
```

cloudmarker/stores/mongodbstore.py
```diff
@@ -0,0 +1,100 @@
...
+
+from pymongo import MongoClient
+
+_log = logging.getLogger(__name__)
+
+
+class MongoDBStore:
+    """A plugin to store records on the MongoDB."""
+
+    def __init__(self, host, port, username=None, password=None, db='gcp',
```
prateeknischal: Depending on the [comment](https://github.com/cloudmarker/cloudmarker/pull/20/commits/9efa467d41b3f51701f5bb509ea3adfa609d3cf1#r250082197), the args list should be something like:
`def __init__(self, username, password, host, port=27017, db='gcp', buffer_size=1000)`
This would allow more cohesiveness with the `__init__` and defaults used by mongo client.

cloudmarker/stores/mongodbstore.py
```diff
@@ -0,0 +1,100 @@
...
+            port (int): MongoDB instance port.
+            username (str): MongoDB instance auth username.
+            password (str): MongoDB instance auth password.
+            db (str): MongoDB database name, default: ``gcp``
+            buffer_size (int): Max number of records to hold in memeory for
+                each record_type.
+
+        """
+        # pylint: disable=too-many-instance-attributes
+
```
prateeknischal: mongodb db connection params: `self._mongodb_host`, `self._mongodb_port`, `self._mongodb_username`, `self._mongodb_password` are not being used anywhere else and are not required, they can be removed from the instance variables.

cloudmarker/stores/mongodbstore.py
```diff
@@ -0,0 +1,100 @@
...
+        else:
+            self._client = MongoClient(host=self._mongodb_host,
+                                       port=self._mongodb_port,
+                                       username=username,
+                                       password=password,
+                                       authSource=self._db,
+                                       authMechanism='SCRAM-SHA-256',
+                                       connect=False)
+
+    def _flush(self, record_type, records):
```
prateeknischal: Is the second param really required, since we already have a reference to the `self._buffer` dict. The `record` param is just a pointer to the actual array, I think it can be simplified by just doing a lookup from the buffer.

prateeknischal: Is the `record` argument required for `_flush`. Since this function is intended to be used within the class and not with custom data and it has access to the `self._buffer` object as well, I think only the key value `record_type` would be necessary and simplify the signature.

cloudmarker/stores/mongodbstore.py
```diff
@@ -0,0 +1,100 @@
...
+
+        # records gets more than buffer size then flush record_type records to
+        # MongoDB
+        if len(records) >= self._buffer_size:
+            self._flush(record_type, records)
+
+    def done(self):
+        """Flush remaining records from the buffer."""
+        # Flush all the residual records buffers to mongodb collections
+        for record_type, records in self._buffer.items():
```
prateeknischal: iterating over a changing dictionary will cause a `RuntimeException` here. Ref: [how-to-avoid-runtimeerror-dictionary-changed-size-during-iteration-error](https://stackoverflow.com/questions/11941817/how-to-avoid-runtimeerror-dictionary-changed-size-during-iteration-error).
Snippet:
```
def done(self):
        for record_type, x in self._buffer.items():
            self._flush(record_type)
```
Response:
```
$ python3 mongo_test.py
Inserted 3 out of 3 documents for record type: users
Traceback (most recent call last):
  File "mongo_test.py", line 24, in <module>
    mongo_store()
  File "mongo_test.py", line 21, in mongo_store
    store.done()
  File "/Users/p0n002h/git/py-orm/mongo/mongostore.py", line 64, in done
    for record_type, x in self._buffer.items():
RuntimeError: dictionary changed size during iteration
```

As recommended, please use `list(self._buffer.items())` in this version

cloudmarker/stores/mongodbstore.py
r0hi7: Connection to MongoDB server may or may not be authenticated. If the default keyword argument is None, then the connect call to mongo will fail with the following error.

```
  File "/Users/r0s00s4/.venv/cloudmarker/lib/python3.7/site-packages/pymongo/client_options.py", line 38, in _parse_credentials
    mechanism, source, username, password, options, database)
  File "/Users/r0s00s4/.venv/cloudmarker/lib/python3.7/site-packages/pymongo/auth.py", line 107, in _build_credentials_tuple
    raise ConfigurationError("%s requires a username." % (mech,))
pymongo.errors.ConfigurationError: SCRAM-SHA-256 requires a username.
``` 
The last line of trace shows the reason for failure to connect the call. In background the PyMongo will create a URI like this `"mongodb://user:password@example.com/?authSource=the_database&authMechanism=SCRAM-SHA-256"` if None is passed, the call would look something like `"mongodb://:@example.com/?authSource=the_database&authMechanism=SCRAM-SHA-256"` which is a malformed URI to connect to.

The alternative to this is to keep the empty string in username and password in the keyword argument.

cloudmarker/stores/mongodbstore.py
r0hi7: `db` is default argument, how to keep default argument before non-default.
I have used the default name for the `db` which is `gcp` so that in case if we plan to integrate this application(or its store that is MongoDB) to some dashboard then it will easily in application control and would require less configurations, if this field is configured by a user of the application (which is possible now as well) then we need some way of keeping track of that name for the dashboard as well.

cloudmarker/stores/mongodbstore.py
susam: @prateeknischal What's the behaviour with empty string value for `username` and `password` when you connect with MongoClient. Does the MongoDB support empty username and password? If it does, then we would have to investigate why the Python API does not support it.

I am yet to experiment with MongoDB, so right now, I am just going to rely on your knowledge to form my own opinion about your other comment related to `MongoDBStore.__init__` parameter order.

s-nirali: The python API does support not having to initialise `username` and `password` as shown in this function definition http://api.mongodb.com/python/current/api/pymongo/database.html#pymongo.database.Database.authenticate. @prateeknischal and I tested it out when I happened to come across this function. Even though it is deprecated, it still means that Mongo DB allows storing without authentication.

prateeknischal: @susam If you set up the mongo DB server without any username or password, then python API accepts both `None` and `""` as both `username` and `password`.

prateeknischal: To test that please update [docker-compose.yml](https://gecgithub01.walmart.com/p0n002h/py-orm/blob/master/mongo/docker-compose.yml) file by removing the `enviroment` sections in both the containers, update `username` and `password` in [mongo_test.py](https://gecgithub01.walmart.com/p0n002h/py-orm/blob/master/mongo/mongo_test.py) file and then run:
```
. venv
make mongo
```

cloudmarker/stores/mongodbstore.py
prateeknischal: Okay, let me present this in a different way. Usually, default host and port for any database is `localhost` and standard port for the database type. And in all cases, you need to give the DB name. You can have all params as default to make it work OOB, but it's not very practical. It is a very common scenario that people are using shared databases, so having a DB name as a required arg does not make config a lot difficult but just a regular config.

cloudmarker/stores/mongodbstore.py
prateeknischal: @r0hi7, just got curious on the URL format. 
`user` and `password` are optional params in the [RFC 1738](https://www.ietf.org/rfc/rfc1738.txt), so `mongodb://:@example.com/?authSource=the_database&authMechanism=SCRAM-SHA-256` is still valid.
```
user
    An optional user name. Some schemes (e.g., ftp) allow the
    specification of a user name.

password
    An optional password. If present, it follows the user
    name separated from it by a colon.
```

susam:
Rebased this pull request on `master` to ensure that it sits neatly on top of the latest work and there are no merge conflicts.

I will begin testing and reviewing this now.

cloudmarker/stores/mongodbstore.py
susam: Created issue #28 to make a decision about this and make the required changes in a separate commit.

PR #21, opened 2019-01-17
Add: Handling ~ in the path for FileStore plugin

### LINKED ISSUE #17: Support tilde expansion on filestore path
We should support [tilde expansion](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_01) in any paths read from configuration as much as possible.

For example, currently the `filestore` plugin module accepts a directory path as `path` parameter to write output files.

https://github.com/cloudmarker/cloudmarker/blob/273eadb325f9183f1932272efb3b0ce55e3bda7a/cloudmarker/stores/filestore.py#L11-L18

It should expand tilde (`~`) or (`~user`) in the `path` to the home directory of the user.

Python has a [`os.path.expanduser()`](https://docs.python.org/3.7/library/os.path.html#os.path.expanduser) function that does this. This should be used to do this task. Here is an example usage:

```
>>> import os
>>> os.path.expanduser('~susam')
'/Users/susam'
>>> os.path.expanduser('~susam/foo')
'/Users/susam/foo'
>>> os.path.expanduser('~')
'/Users/susam'
>>> os.path.expanduser('~/foo')
'/Users/susam/foo'
```

prateeknischal:
Use os.path.expanduser() function to translate the ~ to the home of the
current user.

Add Tests for the above code change

Closes #17

### DIFF
--- a/cloudmarker/test/test_filestore.py
+++ b/cloudmarker/test/test_filestore.py
@@ -1,0 +1,75 @@
+"""Tests for filestore plugin."""
+
+
+import os
+import pathlib
+import shutil
+import unittest
+
+from cloudmarker.stores import filestore
+
+
+def write_to_file(filename):
+    """Write a dummy data to a filename.
+
+    Arguments:
+        filename (str): the file location where dummy data has to be written
+    """
+    # Initialize the FileStore plugin with the filename
+    store = filestore.FileStore(filename)
+
+    # dummy data to write
+    mp = {
+        "record_type": "dummy",
+        "data": {
+            "test_name": "tilde_filename"
+        }
+    }
+
+    store.write(mp)
+    store.done()
+
+
+class FileStoreTest(unittest.TestCase):
+    """Tests for filestore plugin."""
+
+    def test_filepath_tilde_no_user(self):
+        write_to_file("~/caramel_l33t/cloudmarker")
+        # if the parsing fails, it should create a "~" directory in the current
+        # path
+        f = pathlib.Path("~")
+
+        # If exists and is a directory, the parsing failed.
+        self.assertFalse(f.is_dir())
+
+    def test_filepath_tilde_user(self):
+        user = os.getenv("USER")
+        write_to_file("~{}/caramel_tpb/cloudmarker".format(user))
+        # if the parsing fails, it should create a "~" directory in the current
+        # path
+        f = pathlib.Path("~{}".format(user))
+
+        # If exists and is a directory, the parsing failed.
+        self.assertFalse(f.is_file())
+
+    def test_filepath_dir_traversal(self):
+        write_to_file("/tmp/cloudmarker/../abc")
+
+        f = pathlib.Path(os.path.join("/tmp/abc", "dummy.json"))
+        self.assertTrue(f.is_file())
+
+    def tearDown(self):
+        # remove all junk files created after every test
+        user = os.getenv("USER")
+        p = [
+            "/tmp/abc",
+            os.path.expanduser("~/caramel_l33t"),
+            os.path.expanduser("~/caramel_tpb"),
+            "~{}".format(user),
+            "~"
+        ]
+
+        for path in p:
+            if pathlib.Path(path).exists():
+                # delete the directory recursively
+                shutil.rmtree(path)
--- a/cloudmarker/stores/filestore.py
+++ b/cloudmarker/stores/filestore.py
@@ -3,6 +3,7 @@
 
 import json
 import os
+import os.path
 
 
 class FileStore:
@@ -15,9 +16,9 @@ class FileStore:
             path (str): Path of directory where files are written to.
 
         """
-        self._path = path
+        self._path = os.path.expanduser(path)
         self._record_types = set()
-        os.makedirs(path, exist_ok=True)
+        os.makedirs(self._path, exist_ok=True)
 
     def write(self, record):
         """Write JSON records to the file system.


cloudmarker/test/test_filestore.py
```diff
@@ -0,0 +1,69 @@
...
+        store.done()
+
+    def test_filepath_tilde(self):
+        self.write_to_file("~/caramel_l33t/cloudmarker")
+        # if the parsing fails, it should create a "~" directory in the current
+        # path
+        f = pathlib.Path("~")
+
+        # If exists and is a directory, the parsing failed.
+        self.assertFalse(f.exists() and f.is_dir())
```
susam: I think only one of`f.is_dir()` and `f.exists()` should be enough. I don't see any specific advantage of using both checks. What do you think?

cloudmarker/test/test_filestore.py
```diff
@@ -0,0 +1,69 @@
...
+import unittest
+
+from cloudmarker.stores import filestore
+
+
+class FileStoreTest(unittest.TestCase):
+    """Tests for filestore plugins."""
+
+    @classmethod
+    def write_to_file(cls, filename):
```
susam: Since the `cls` argument is never used, it is technically not a class method. It does not work on the class. This could be a static method (which in Python is just syntactic sugar). A much simpler and obvious way to write this function would be as a module level function.

I almost always prefer a module level function to a static method function.

cloudmarker/test/test_filestore.py
```diff
@@ -0,0 +1,69 @@
...
+    def test_filepath_tilde(self):
+        self.write_to_file("~/caramel_l33t/cloudmarker")
+        # if the parsing fails, it should create a "~" directory in the current
+        # path
+        f = pathlib.Path("~")
+
+        # If exists and is a directory, the parsing failed.
+        self.assertFalse(f.exists() and f.is_dir())
+
+    def test_filepath_tilde2(self):
```
susam: We need to make the method names more descriptive so that when a test method fails, we get a fair idea of which functionality failed from the method name itself. For example, this method name could be `test_filepath_tilde_user` and the previous method could be named `test_filepath_tilde_no_user`.

cloudmarker/test/test_filestore.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertFalse(f.exists() and f.is_dir())
+
+    def test_filepath_dir_traversal(self):
+        self.write_to_file("/tmp/cloudmarker/../abc")
+
+        f = pathlib.Path("/tmp/abc")
+        self.assertTrue(f.exists() and f.is_dir())
+
+    @classmethod
+    def tearDownClass(cls):
```
susam: This should be an instance method, i.e., `tearDown(self)`.

Rationale: Every test should run in a clean environment. There should be no coupling between the tests. A `tearDown(self)` call after test method is run reduces the chance of decoupling.

When we see a `tearDownClass(cls)` class method, the reader of the test code has to spend some additional time reading the code of each test to ensure that one method does not inadvertently depend on artifacts left by another test. But if we see a `tearDown(self)` method instead, we know for sure that the environment is getting cleaned up after every test method runs, so if a test method inadvertently depends on artifacts left by another test, then that test method is likely to fail because a `tearDown(self)` call after every test would blow away those artifacts.

cloudmarker/test/test_filestore.py
```diff
@@ -0,0 +1,69 @@
...
+        p = [
+            "/tmp/abc",
+            "~/caramel_l33t",
+            os.path.join("~{}".format(user), "caramel_l33t")
+        ]
+
+        for path in p:
+            if pathlib.Path(path).exists():
+                # delete the directory recursively
+                shutil.rmtree(path)
```
s-nirali: This command did not delete the directories created for me.

cloudmarker/test/test_filestore.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertTrue(f.exists() and f.is_dir())
+
+    @classmethod
+    def tearDownClass(cls):
+        # remove all junk files created
+        user = os.getenv("USER")
+        p = [
+            "/tmp/abc",
+            "~/caramel_l33t",
+            os.path.join("~{}".format(user), "caramel_l33t")
```
s-nirali: There should be one more directory here `caramel_tpb` as in the 2nd test method. And why caramel actually?

cloudmarker/test/test_filestore.py
```diff
@@ -0,0 +1,69 @@
...
+    def test_filepath_tilde2(self):
+        user = os.getenv("USER")
+        self.write_to_file("~{}/caramel_tpb/cloudmarker".format(user))
+        # if the parsing fails, it should create a "~" directory in the current
+        # path
+        f = pathlib.Path("~{}".format(user))
+
+        # If exists and is a directory, the parsing failed.
+        self.assertFalse(f.exists() and f.is_dir())
+
```
s-nirali: I think it would be a more accurate to check for the existence of the file itself which would automatically check the whole path also. Something like `os.path.isfile("$HOME/caramel_tpb/cloudmarker/dummy.json"`. I know that the $HOME is only for Linux, might need something that would work for cross-platform paths. The condition here is minimalistic and beautiful in that it checks if the path was created in the wrong place, but it is not checking if the path expansion actually happened. Any views @prateeknischal , @susam ?

susam: @s-nirali I agree. Ideally, there should be additional tests to check the positive condition, i.e., that the file is written to the correct path. Such a test should be another separate test method. While such a test case would be nice, even if it is missing in this pull request (PR), we can still merge this PR with the current set of test methods after the review comments are addressed, and further tests may be added later by anyone.

By the way, a cross-platform way to get the home directory is `os.path.expanduser('~')`.

cloudmarker/test/test_filestore.py
susam: Looks like the `~` in the path needs to undergo tilde expansion so that the existence check succeeds and the folder is deleted. There is `pathlib.Path(path).expanduser()` but it was introduced in version 3.5, so using this would cause Travis CI builds to fail because we have Python 3.4 also in [the build matrix](https://github.com/cloudmarker/cloudmarker/blob/955d4ef/.travis.yml#L3-L5). So that leaves us with `os.path.exists(os.path.expanduser(path))`.

cloudmarker/test/test_filestore.py
prateeknischal: I had used `f.exists()` and `f.is_dir()` because I did not see that errors are propagated. [Path.is_dir](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_dir) so there is no need to check for `exists()`, `is_dir()` implies `exists()`. I'll move to `is_dir()`.

cloudmarker/test/test_filestore.py
prateeknischal: The [`tearDownClass`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDownClass) method is related to the `unittest` class and is marked to be called after the completion of all the tests in that particular class. The library docs say that it should be marked as a `@classmethod`.

cloudmarker/test/test_filestore.py
prateeknischal: The [`tearDownClass`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDownClass) method is related to the `unittest` class and is supposed to be called after the completion of all the tests in that particular class. The library docs say that it should be marked as a `@classmethod`. If I used `self`, the linter says:
```
cloudmarker/test/test_filestore.py:62:4: C0202 Class method tearDownClass should have 'cls' as first argument [pylint]
```
The reason for not using `teadDown(cls)` is that it runs the same thing after every test. Since every test will generate a different directory, it seems to me that there should be one tearDown at the end which is `tearDownClass`.

cloudmarker/test/test_filestore.py
prateeknischal: Agreed, I had missed that folder.
`caramel` is a random string came to my mind at the time since I thought this test should not delete any actual folders. To make it more random, I added another suffix `tpb` which is the shorthand for [The Pirate Bay](https://en.wikipedia.org/wiki/The_Pirate_Bay) or more appropriately (the actual source) [TPB AFK (2013)](https://www.imdb.com/title/tt2608732/).

cloudmarker/test/test_filestore.py
prateeknischal: I have added the following folder names, so now it should delete all folder with and without path expansion.
```python
user = os.getenv("USER")
p = [
    "/tmp/abc",
    os.path.expanduser("~/caramel_l33t"),
    os.path.expanduser("~/caramel_tpb"),
    "~{}".format(user),
    "~"
]
```

cloudmarker/test/test_filestore.py
susam: The `tearDownClass(cls)` method should be written with `@classmethod` and `cls` because it is a classmethod. This is alright.

I am suggesting that we do not use `tearDownClass(cls)` at all unless there is a very good reason to do so and we use `tearDown(self)` instead because it is run after every test. That way, we know that every test runs in a clean environment (to the extent possible). I offered decoupling of tests as the rationale behind this.

By decoupling of tests, I mean, we should not create a situation where it is possible for `test_method1(self)` to create an artifact that `test_method2(self)` uses. If `test_method2(self)` requires some artifacts for itself to function, then it should itself create those artifacts or they should be created just before the test method runs via `setUp(self)` (not to be confused with `setUpClass(self)`).

If we perform a tear down only at the end of all tests in a test class, then we might inadvertently (not now but sometime in future) create a file in `test_method1(self)` that `test_method2(self)` uses and we would not know about it if we are using `tearDownClass(cls)`. The tests would run fine and we would not know that we have coupled the two test methods. If, however, we use `tearDown(self)`, then `test_method2(self)` would fail because the artifacts it is inadvertently depending on would have been removed by `tearDown(self)`. After seeing this failure, we would know that we need to fix `test_method2(self)` and decouple it from `test_method1(self)`.

Having said that, you should address this comment only if you agree with the rationale. It is okay not to address this comment as well. We can merge this pull request with or without addressing this review comment because the test code is correct in either case.

cloudmarker/test/test_filestore.py
prateeknischal: I have added `pathlib.Path(os.path.join("/tmp/abc", "dummy.json"))` for testing the positive condition, but as discussed with @s-nirali, there is actually no need for this test file as we are now basically testing python code and not cloudmarker code.

To write a positive test for expanduser, I will have to use expanduser function, which does not make any sense. Would it be ok to remove all the tests and let just the have the test in #22 only.

susam: I think there should be at least one test for this. The fact that `FileStore` performs tilde expansion on the path is a Cloudmarker behavior. If someday in future, we inadvertently stop doing tilde expansion, then at least one test should break that should tell us that the expected behavior has gone missing.

Regarding invoking `os.path.expanduser()` to test something that does `os.path.expanduser()`, I agree that it seems self-referential indeed. However, at the same time it is also good to capture the expected behaviour as a test. At least, it would allow for things like changing the `FileStore` implementation from using `os.path.expanduser()` to something like `pathlib.Path(path).expanduser()` in future, say we don't want to support Python 3.4 anymore. The test would guarantee that nothing breaks during such a change.

This turns out to be one of those decisions that depends on one's judgement. I prefer having at least one test. I'll let you decide what you want to do with the tests. 🙂

By the way, the `pathlib` module is featureful enough to not require `os.path.join()`. For example, `pathlib.Path('/tmp/abc', 'dummy.json')` produces the same result as `pathlib.Path(os.path.join("/tmp/abc", "dummy.json"))`.

PR #27, opened 2019-01-29
Generate documentation for firewallcheck module

s-nirali:


susam:
Please change the commit summary to something like this:

    Generate documentation for firewallcheck module

I think "Generate" more precisely tells us that we are modifying the Sphinx project to generate the documentation for the `firewallcheck` module rather than updating the docstrings within the module.

Please amend the commit, rebase it on top of the latest master, and force-push it.

PR #32, opened 2019-02-02
MongoDBStore enhancements: arbitrarily ordered insertion, parameter list, single MongoClient statement

prateeknischal:
Updated the MongoDBStore Plugin initialization params sequence to make it more intuitive.
1. `db`: name of the database, as this is the natural level of separation in a shared DB
2. `username`: For a shared DB, username is required
3. `password`: Required Authentication
4. `host`: defaults to localhost, but should be required only after authentication information if it's a remote DB
5. `port`: default 27017 port for localhost, or an specific port 
6. `buffer_size`: As this is just a buffering param, can be left to default

Change the mode of `insert_many` to `ordered=False`, as the default behavious allows the driver to insert the documents in order and fail at the first failure. By making it False, the driver will try to insert data in an unordered format and will try insertion for all documents at least once, possibly in parallel.
[insert_many](http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.insert_many).

To handle the failed insertions due to error, surround the `insert_many` with try-catch and log the incident.

Close the connection when `done` is called.

### DIFF
--- a/cloudmarker/stores/mongodbstore.py
+++ b/cloudmarker/stores/mongodbstore.py
@@ -11,19 +11,22 @@ _log = logging.getLogger(__name__)
 class MongoDBStore:
     """A plugin to store records on MongoDB."""
 
-    def __init__(self, host, port, username=None, password=None, db='gcp',
+    def __init__(self, db, username, password, host, port=27017,
                  buffer_size=1000):
         """Create an instance of :class:`MongoDBStore` plugin.
 
-        Arguments:
-            host (str): MongoDB instance host.
-            port (int): MongoDB instance port.
-            username (str): MongoDB instance auth username.
-            password (str): MongoDB instance auth password.
-            db (str): MongoDB database name, default: ``gcp``
-            buffer_size (int): Max number of records to hold in memeory for
-                each record_type.
+        It will use the default port for mongodb 27017 if not specified.
+        The Authentication scheme will be negotiated by MongoDB and the client
+        for v4.0+ to SCRAM-SHA-1 or SCRAM-SHA-256 by default aftere
+        negotiation.
 
+        Arguments:
+            db (str): name of the database
+            username (str): username for the database
+            password (str): password for username to authenticate with the db
+            host (str): hostname for the DB server
+            port (int): port for mongoDB is listening, defaults to 27017
+            buffer_size (int): max buffer before flushing to db
         """
         # pylint: disable=too-many-instance-attributes
 
@@ -33,8 +36,8 @@ class MongoDBStore:
         self._mongodb_password = password
         self._db = db
 
-        self._buffer_size = buffer_size
         self._buffer = {}
+        self._buffer_size = buffer_size
 
         if not (username and password):
             self._client = MongoClient(host=self._mongodb_host,
--- a/cloudmarker/stores/mongodbstore.py
+++ b/cloudmarker/stores/mongodbstore.py
@@ -3,7 +3,7 @@
 
 import logging
 
-from pymongo import MongoClient
+from pymongo import MongoClient, errors
 
 _log = logging.getLogger(__name__)
 
@@ -60,11 +60,17 @@ class MongoDBStore:
         _log.info('Inserting %i documents into collection %s ...',
                   len(records), record_type)
 
-        self._client[self._db][record_type].insert_many(records)
+        try:
+            # Making the bulk insert ordered=False, to make sure that all
+            # of the documents are tried to be inserted. In the default
+            # config, it will fail after the first error
+            res = self._db[record_type].insert_many(records, ordered=False)
+            _log.info('Saved %i out of %i documents into collection: %s',
+                      len(res.inserted_ids), len(records), record_type)
 
-        _log.info('Saved %i documents into collection %s',
-                  len(records), record_type)
-        del records[:]
+            del records[:]
+        except errors.BulkWriteError as bwe:
+            _log.error('Failed to write records with error %s', bwe)
 
     def write(self, record):
         """Write JSON records to the MongoDB collections.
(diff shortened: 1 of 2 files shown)


cloudmarker/stores/mongodbstore.py
```diff
@@ -3,51 +3,67 @@
...
+                validation fails. ``error`` will raise an error while ``warn``
+                will log that violation in the db server logs.
+        """
+        for model in kwargs.get('models', []):
+            if model.collection() not in self._db.list_collection_names():
+                self._db.create_collection(
+                    model.collection(),
+                    validator=model.validator(),
+                    validationAction=model.enforce()
+                )
```
susam: I think this validation is not required for the problem we are trying to solve. We want all data to be entered into our data stores regardless of how they look. As long as the data is valid JSON, we enter it into our data stores. The reason is that we do not have control on the data sources. Whatever data is presented to us by the cloud provider is the data we have to save and work with. Further, in a security software like this usually all data needs to be saved regardless of whether they adhere to a specific schema or not.

My opinion is that this change should not be merged into `master`. I would like to know if you or anyone else wants this to be merged. Since this is a community project and we are a developer community of four right now, I think we can take a decision based on what the community thinks.

By the way, this is one reason why separate concerns should be separate pull requests. Right now, there are two commits in this pull request addressing separate concerns and if we decide not to merge this, then the whole pull request will remain blocked until this change is removed. Therefore, I request you to make each pull request further smaller in scope so that each can be reviewed and merged independently. At the very least, this validation change should be moved out to a separate pull request so that it can be discussed separately.

cloudmarker/stores/mongodbstore.py
```diff
@@ -3,51 +3,67 @@
...
+        following functions:
+            collection(): returns the name of the collection
+            validator(): returns a valid string with $jsonSchema that contains
+                all the validation properties as specified in the mongodb
+                docs.
+            enforce(): returns one of the strings ``error`` or ``warn``.
+                This will be used to define the error level for when the schema
+                validation fails. ``error`` will raise an error while ``warn``
+                will log that violation in the db server logs.
+        """
```
susam: If it turns out that this change should indeed be merged, then the above comment needs to be reformatted with regular `#`-prefixed comments.

We reserve the usage of `"""`-quoted comments only for docstrings.

cloudmarker/stores/mongodbstore.py
```diff
@@ -40,6 +40,27 @@ def __init__(self, db, username, password, host, port=27017,
...
+        for model in kwargs.get('models', []):
+            if model.collection() not in self._db.list_collection_names():
+                self._db.create_collection(
+                    model.collection(),
+                    validator=model.validator(),
+                    validationAction=model.enforce()
+                )
```
susam: If it turns out that this change should be merged to `master`, please update the commit message to:

    Add optional validations for MongoDB schema

In general, the commit messages need to be very short (not exceeding 50 characters in length) but very precise with just enough detail to tell us what exactly is going on in the change without compromising on the way we write brand names ("MongoDB" in this case).

cloudmarker/stores/mongodbstore.py
```diff
@@ -60,11 +76,17 @@ def _flush(self, record_type, records):
...
                   len(records), record_type)
 
-        self._client[self._db][record_type].insert_many(records)
+        try:
+            # Making the bulk insert ordered=False, to make sure that all
+            # of the documents are tried to be inserted. In the default
+            # config, it will fail after the first error
+            res = self._db[record_type].insert_many(records, ordered=False)
+            _log.info('Saved %i out of %i documents into collection: %s',
+                      len(res.inserted_ids), len(records), record_type)
```
susam: The `ordered` parameter being `True` by default is a good catch. Thank you for making this change.

cloudmarker/stores/mongodbstore.py
```diff
@@ -104,3 +126,7 @@ def done(self):
         # Flush all the residual records buffers to mongodb collections
         for record_type, records in self._buffer.items():
             self._flush(record_type, records)
+
+        # Close the mongo db connection when done is called as this will be the
```
susam: Replace "mongo db" with "MongoDB".

cloudmarker/stores/mongodbstore.py
```diff
@@ -3,51 +3,46 @@
...
+            username (str): username for the database
+            password (str): password for username to authenticate with the db
+            host (str): hostname for the DB server
+            port (int): port for mongoDB is listening, defaults to 27017
+            buffer_size (int): max buffer before flushing to db
+            kwargs (dict): Additional args
+                * models - List of classes for validator and enforcement
+                    requirements.
         """
-        # pylint: disable=too-many-instance-attributes
```
susam: Removal of `pylint` directive has to be a separate commit. Every commit should do exactly what the commit summary line says and no more. Further, the commit summary line cannot be generic phrases "code cleanup" or "error handling". 🙂

This single line change should be a separate commit (could be a separate pull request or part of any pull request) that clearly tells in the summary line, for example, `Remove Pylint control directive from MongoDBStore`. Worth noting in this example:

- The summary line is very specific about what change is made.
- The brand names "Pylint" is capitalized just the way the Pylint community has named their project.

Then the commit message should elaborate why it is okay to remove this `Pylint` directive. I believe the answer is that `pylama.ini` already has the necessary `ignore` keys which makes this directive redundant.

cloudmarker/stores/mongodbstore.py
prateeknischal: The reason I want this feature is to allow basic validation into fields, say some `id`, or some `name` field. This would create one safeguard if someone tries to push something incorrect or non-relevant into this collection as this is just a collection level check.

cloudmarker/stores/mongodbstore.py
prateeknischal: Fixed: [updated to MongoDB](https://github.com/cloudmarker/cloudmarker/pull/32/files#diff-508ea730766e9a753ad4ce9b65ec1da4R105)

PR #33, opened 2019-02-02
MongoDB Store Plugin Integration Tests

prateeknischal:
This commit adds a `docker-compose.yml` file to spin up a mongoDB server and UI for automated
tests and inspection.

The docker file is configured to use the same credentials as the tests.
Use `docker-compose -f cloudmarker/test/data/docker-compose.yml up` to
run the stack.

Add follwing tests:
- happy path test
- valid insert
- invalid insert on schema validation
- insert with warning on schema validation

### DIFF
--- a/cloudmarker/stores/mongodbstore.py
+++ b/cloudmarker/stores/mongodbstore.py
@@ -46,13 +46,13 @@ class MongoDBStore:
         # done with the following way. The object specified must have the
         # following functions:
         #     collection(): returns the name of the collection
-        #     validator(): returns a valid string with $jsonSchema that contains
-        #         all the validation properties as specified in the mongodb
-        #         docs.
+        #     validator(): returns a valid string with $jsonSchema that
+        #         contains all the validation properties as specified in the
+        #         mongodb docs.
         #     enforce(): returns one of the strings ``error`` or ``warn``.
-        #         This will be used to define the error level for when the schema
-        #         validation fails. ``error`` will raise an error while ``warn``
-        #         will log that violation in the db server logs.
+        #         This will be used to define the error level for when the
+        #         schema validation fails. ``error`` will raise an error while
+        #         ``warn`` will log that violation in the db server logs.
         for model in kwargs.get('models', []):
             if model.collection() not in self._db.list_collection_names():
                 self._db.create_collection(
--- a/cloudmarker/stores/mongodbstore.py
+++ b/cloudmarker/stores/mongodbstore.py
@@ -12,7 +12,7 @@ class MongoDBStore:
     """A plugin to store records on MongoDB."""
 
     def __init__(self, db, username, password, host, port=27017,
-                 buffer_size=1000):
+                 buffer_size=1000, **kwargs):
         """Create an instance of :class:`MongoDBStore` plugin.
 
         It will use the default port for mongodb 27017 if not specified.
@@ -27,6 +27,9 @@ class MongoDBStore:
             host (str): hostname for the DB server
             port (int): port for mongoDB is listening, defaults to 27017
             buffer_size (int): max buffer before flushing to db
+            kwargs (dict): Additional args
+                * models - List of classes for validator and enforcement
+                    requirements.
         """
         self._client = MongoClient(
             host=host,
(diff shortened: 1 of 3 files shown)


susam:
Thank you for making this a separate pull request. I will start reviewing this after #32 is taken care of.

cloudmarker/test/data/docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+    ports:
+      - 8081:8081
+    environment:
+      ME_CONFIG_MONGODB_ADMINUSERNAME: dummy
+      ME_CONFIG_MONGODB_ADMINPASSWORD: password
+    networks:
+      mynet:
+        ipv4_address: 172.16.101.3
+
+networks:
```
r0hi7: Why are you choosing a specific subnet for the containers? what I feel is like, we should let the docker spin up the network that it wants to and attach the containers to it. I could not see any use of adding network here. 
Is it that networks are mandatory in compose version `3.1` and if it so do we have a simpler way of referencing the container instead of static IPs.

cloudmarker/test/data/docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+      MONGO_INITDB_ROOT_USERNAME: dummy
+      MONGO_INITDB_ROOT_PASSWORD: password
+    ports:
+      - 27017:27017
+    networks:
+      mynet:
+        ipv4_address: 172.16.101.2
+
+  mongo-express:
+    image: mongo-express
```
r0hi7: I can see mongo-express is added only for a testing purpose of MongoDB store. I could not figure out any other use of this image. Is it okay to take this image into docker compose @susam as in future we need to maintain mongo-express separately only for testing the MongoDB store plugin.

cloudmarker/test/data/docker-compose.yml
```diff
@@ -0,0 +1,35 @@
+# Use root/example as user/password credentials
```
r0hi7: I think its okay to remove this comment as this is misleading to me at least.

cloudmarker/test/test_mongostore.py
```diff
@@ -0,0 +1,205 @@
...
+    def test_connection(self):
+        client = MongoClient(host='localhost', port=27017,
+                             username='dummy', password='password')
+
+        db = client["test"]
+
+        for collection in db.list_collection_names():
+            db[collection].drop()
+
+        store = mongodbstore.MongoDBStore('test', 'dummy',
```
r0hi7: Use keywords argument for more clarity.

cloudmarker/stores/mongodbstore.py
```diff
@@ -3,51 +3,66 @@
...
+        #     enforce(): returns one of the strings ``error`` or ``warn``.
+        #         This will be used to define the error level for when the
+        #         schema validation fails. ``error`` will raise an error while
+        #         ``warn`` will log that violation in the db server logs.
+        for model in kwargs.get('models', []):
+            if model.collection() not in self._db.list_collection_names():
+                self._db.create_collection(
+                    model.collection(),
+                    validator=model.validator(),
+                    validationAction=model.enforce()
```
r0hi7: Why are you planning to add model validators here? We only fetch the data and keep it in the DB as it.

cloudmarker/test/test_mongostore.py
r0hi7: @prateeknischal & @susam I have a curiosity about this test case. Why do we have an integration test case for this store plugin as currently, we don't have any CD in the pipeline? I am not pretty much sure if this integration test would serve any purpose and if it does what would it be. I was thinking if we can have a mock unit test to test the MongoDB store connect would fit in this scenario. Also is it that we are trying to test the pymongo API here? 
I would like to know your thoughts on this @susam.

susam: In general, it is good to have integration tests. As the project becomes more and more complex, integration tests would tell us if end-to-end functionality is running fine or not. At the same time, I think it is too soon to be writing integration tests right now.

Developing integration tests pose a many challenges: For example, how do we connect to a test instance of GCP only to perform integration tests? Where would we get such a test instance from? All these details have to be worked out. Integration tests also increases the build time, so we have to figure out a way to run integration tests less frequently only on certain builds (like on important milestones or before releases). Integration tests also add _a lot_ of complexity to the tests. Do we as a team enough free time to maintain these integration tests?

For stores based on open source components, it is a little easier to do integration tests. We can spin up Docker containers for MongoDB, Elasticsearc, etc. and test against these containers. But concerns about increased build times and maintenance overhead remain.

As far as unit testing with mock is concerned, you are right. That is the right way to do unit testing. I will spend sometime on exploring if we can use `mock.patch` to unit-test the stores without having to spin up actual instances of MongoDB, Elasticsearch, etc.

Our first priority is unit testing and better code coverage using unit tests. Integration tests is low priority right now. Since integration tests would be done less frequently, integration tests also won't count towards the code coverage metrics that are run on every build.

cloudmarker/test/data/docker-compose.yml
susam: I am curious to know where this `mongo-express` image is being used.

susam:
Closing this pull request for now. This is quite old right now and not being worked upon. We have mostly settled on mocking external dependencies with `unittest.mock` during unit tests. In case, anyone is interested to pick the integration test work again, please reopen this pull request or create a new pull request for it.

PR #38, opened 2019-02-08
Add unit tests for MongoDBStore

susam:
The unit tests use `mock.patch` to patch `MongoClient` with `MagicMock` objects in order to avoid dependency on an actual MongoDB server to run the tests.

Note that in the tests we do not call a `Mock` object's `assert_called` method although this method would have sufficed. That's because `assert_called` method was introduced in Python 3.6. We support Python versions 3.4 and 3.5 also at this time and therefore run Travis CI builds with these versions. Calling `assert_called()` leads to test error with Python 3.5 due to `AttributeError` and does not do what we want on Python 3.4 (returns a `MagicMock` object that does nothing).

In order to support Python versions 3.4 and 3.5, we use `assert_called_with` and `assert_called_once_with` methods because these methods are available prior to Python 3.6.

### DIFF
--- a/cloudmarker/test/test_mongodbstore.py
+++ b/cloudmarker/test/test_mongodbstore.py
@@ -1,0 +1,123 @@
+"""Tests for mongodbstore plugin."""
+
+
+import unittest
+from unittest import mock
+
+from cloudmarker.stores import mongodbstore
+
+
+class MongoDBStoreTest(unittest.TestCase):
+    """Tests for mongodbstore plugin."""
+
+    @mock.patch('cloudmarker.stores.mongodbstore.MongoClient')
+    def test_write_without_flush(self, mock_client):
+        # Set buffer size of 3.
+        store = mongodbstore.MongoDBStore(buffer_size=3)
+
+        # But insert only 2 records.
+        for i in range(2):
+            record = {'record_type': 'foo', 'i': i}
+            store.write(record)
+
+        # Verify that buffer is not being flushed via client.
+        mock_method = mock_client()['fake_db']['fake_record_type'].insert_many
+        mock_method.assert_not_called()
+
+    @mock.patch('cloudmarker.stores.mongodbstore.MongoClient')
+    def test_write_with_flush(self, mock_client):
+        # Set buffer size of 3.
+        store = mongodbstore.MongoDBStore(buffer_size=3)
+
+        # Then insert 3 records.
+        for i in range(3):
+            record = {'record_type': 'foo', 'i': i}
+            store.write(record)
+
+        # Verify that buffer is being flushed via client.
+        mock_method = mock_client()['fake_db']['fake_record_type'].insert_many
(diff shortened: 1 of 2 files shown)


susam:
A mysterious issue came up while working on this pull request. An earlier draft of this pull request used the [`assert_called()`](https://docs.python.org/3.7/library/unittest.mock.html#unittest.mock.Mock.assert_called) method of `Mock`. This caused the tests to fail with Python 3.5 on Travis CI which is expected because the documentation for this method clearly states:

> New in version 3.6.


#### Mystery

However, the tests passed successfully on Python 3.4. How could the tests pass in Python 3.4 when this method was introduced only in Python 3.6?


#### Answer

- Python 3.4: `Mock().assert_called()` works because we can call or access any attribute or method (even non-existent ones) of a mock. That's the whole purpose of a mock. We can invoke any non-existent method of a mock and the mock object would keep returning us mock objects, so the calls never fail. So `Mock().foo`, `Mock().bar()`, `Mock().nonsense.nonsense()`, etc. all return mock objects.
- Python 3.5: The Python 3.4 behaviour has a subtle issue. If we invoke `Mock().assert_called()` thinking that this is a unit-testing assert method, we get unexpected results. Since `Mock().assert_called()` only returns a mock in Python 3.4, the call never fails. No assertion is done of course because this is a non-existent method. Calling `Mock().assert_called()` in Python 3.4 would mislead us to believe that our unit tests are good. Therefore, since Python 3.5, there is an explicit check in the `mock` module that raises `AttributeError` if we call a method whose name begins with `assert` on `Mock()`. [[Source code](https://github.com/python/cpython/blob/3.5/Lib/unittest/mock.py#L584-L585)]
- Python 3.6:  An assert method named `Mock().assert_called()` is introduced in Python 3.6, so we can use it safely. [[Source code](https://github.com/python/cpython/blob/3.6/Lib/unittest/mock.py#L779-L786)]


#### Evidence from Debug Builds

- Python 3.4: https://travis-ci.com/cloudmarker/cloudmarker/jobs/176345652#L5350 (`assert_called` leads to creation of a `MagicMock` on the fly)
- Python 3.5: https://travis-ci.com/cloudmarker/cloudmarker/jobs/176345653#L5418 (`assert_called` leads to `AttributeError` due to explicit check)
- Python 3.6: https://travis-ci.com/cloudmarker/cloudmarker/jobs/176345654#L5456 (`assert_called` is a genuine `method` meant for assertions)


#### Evidence from Changelog ###

- Python 3.5: https://docs.python.org/3/whatsnew/3.5.html#unittest-mock:

> The class constructor has a new unsafe parameter, which causes mock objects to raise `AttributeError` on attribute names starting with `"assert"`.

- Python 3.6: https://docs.python.org/3/whatsnew/3.6.html#unittest-mock:

> Two new methods, `Mock.assert_called()` and `Mock.assert_called_once()` to check if the mock object was called.

PR #42, opened 2019-02-25
Fix docstring indentation in MongoDBStore

s-nirali:
The commit fixes the following warning that occurs while running
`make docs`.

    cloudmarker/stores/mongodbstore.py:docstring of
    cloudmarker.stores.mongodbstore.MongoDBStore:24:
    WARNING: Unexpected indentation.

### DIFF
--- a/cloudmarker/stores/mongodbstore.py
+++ b/cloudmarker/stores/mongodbstore.py
@@ -29,7 +29,7 @@ class MongoDBStore:
             buffer_size (int): maximum number of records to buffer
             kwargs (dict): Additional args
                 * models - List of classes for validator and enforcement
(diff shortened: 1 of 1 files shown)


s-nirali:
Should I instead just create an issue to track this so you can fix it while fixing any other docstring issues in a single commit and close this pull request without merging?

susam:
We can merge this now itself. I don't think there would be a single commit that would take care of all docstring fixes. As we write more docstrings, there would be more errors in docstrings, and there would be more fixes for them. Since maintaining good docstrings is a long term sustained work, we will keep merging docstring fixes as soon as we have them.

PR #43, opened 2019-02-25
Handle cloud errors in AzureCloud

s-nirali:
Some subscriptions in Azure could result in the following exception at line [123](https://github.com/cloudmarker/cloudmarker/pull/43/commits/aa82f0a272ff063c23ee019c52cb6e739dcf76c2#diff-4fb06491b278fdc6193c4b1e9b8d0284R123) in cloudmarker/clouds/azurecloud.py while iterating over the objects of any specified Azure provider.

```
msrestazure.azure_exceptions.CloudError: Azure Error: DisallowedOperation
Message: The current subscription type is not permitted to perform operations on any provider namespace. Please use a different subscription.
```

This commit aims to catch these exceptions and handle them gracefully by letting the code continue reading documents from other Azure subscriptions instead of exiting when any subscription throws the above exception.

### DIFF
--- a/cloudmarker/clouds/azurecloud.py
+++ b/cloudmarker/clouds/azurecloud.py
@@ -13,6 +13,7 @@ from azure.mgmt.compute import ComputeManagementClient
 from azure.mgmt.network import NetworkManagementClient
 from azure.mgmt.resource import ResourceManagementClient, SubscriptionClient
 from azure.mgmt.storage import StorageManagementClient
+from msrestazure.azure_exceptions import CloudError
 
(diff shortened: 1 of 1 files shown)


susam:
Correction to my previous comment: Please keep the new commit in this pull request rebased on top of `fixindent` like it is right now. That way I can merge #42 first and then I can merge this (#43) as it was originally intended by you.

s-nirali:
Updated the commit message.

s-nirali:
Rebased over the latest master.

Issue #47, opened 2019-03-10
Place closing square bracket in its own line in the FileStore output files


susam:
Right now, the output of `FileStore` at say, looks like this:

```
[
{
  "raw": {
    "data": 0
  },
  ...
  "com": {
    "type": "mock",
    "origin_worker": "mockaudit-mockcloud",
    "origin_type": "cloud",
    "cloud_worker": "mockaudit-mockcloud",
    "store_worker": "mockaudit-filestore"
  }
}]
```

The opening square bracket is in its own line but the closing square bracket is not. Place the square bracket in its own line, so that the output looks symmetrical.

PR #55, opened 2019-03-23
Add EmailStore plugin

sunnysharmagts:
- Wrote a basic implementation of the emailstore for sending anomalies
  alerts via email. It reads the mail paramaters from the
config.base.yaml file.

### DIFF
--- a/config.base.yaml
+++ b/config.base.yaml
@@ -24,6 +24,16 @@ alerts:
       to:
         - foo@example.com
       text: foo
+  emailalert:
+    plugin: cloudmarker.alerts.emailalert.EmailAlert
+    params:
+      host: smtp.example.com
+      port: 25
+      sender: CloudMarker Alert <no-reply@example.com>
+      to:
+        - recipient@example.com
+      subject: Anomaly notification
+      body: default text
 
 audits:
   mockaudit:
@@ -35,10 +45,10 @@ audits:
       - mockevent
     alerts:
       - filestore
+      - emailalert
 
 run:
   - mockaudit
-
 
 logger:
 
@@ -73,6 +83,5 @@ logger:
     handlers:
       - console
       - file_handler
-
 
 schedule: "00:00"
--- a/docs/api/cloudmarker.stores.rst
+++ b/docs/api/cloudmarker.stores.rst
@@ -9,6 +9,14 @@ cloudmarker.stores package
 Submodules
 ----------
 
+cloudmarker.stores.emailalert module
+------------------------------------
+
+.. automodule:: cloudmarker.stores.emailalert
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
 cloudmarker.stores.filestore module
 -----------------------------------
 
--- a/pylama.ini
+++ b/pylama.ini
@@ -19,8 +19,11 @@ ignore = E1101
 # E1101 Instance of 'Resource' has no 'instances' member [pylint]
 
 [pylama:cloudmarker/stores/mongodbstore.py]
-ignore = R0913 
+ignore = R0913
 # R0913 Too many arguments (7/5)
+
+[pylama:cloudmarker/alerts/emailalert.py]
+ignore = R0913,C0412
 
 [pylama:cloudmarker/stores/splunkhecstore.py]
 ignore = R0913,W0703
--- a/cloudmarker/alerts/emailalert.py
+++ b/cloudmarker/alerts/emailalert.py
@@ -1,0 +1,77 @@
+"""Email alert plugin."""
+
+
+import email
+import logging
+import smtplib
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+
+# Define module-level logger.
+_log = logging.getLogger(__name__)
+
+
+class EmailAlert:
+    """A plugin to send email notification for anomalies found."""
+
+    def __init__(self, host, port, subject, to, sender, body):
+        """Create an instance of :class:`EmailAlert` plugin.
+
+        Arguments:
+            host (string): Server hostname for SMTP.
+            port (int): Server port for SMTP.
+            subject (string): Subject line for email.
+            to (list): List of recipients (string).
+            sender (string): From field for email.
+            body (string): Email body.
+
+        """
+        self.host = host
+        self.port = port
+        self.subject = subject
+        self.to = to
+        self.sender = sender
+        self.body = body
+        self.stringbuffer = []
+
+    def write(self, record):
+        """Write JSON records to the file system.
+
+        This method is called once for every ``record`` read from a
+        cloud. In this example implementation of a alert, we simply
+        send the ``record`` in JSON format via email to the recipient.
+        The records keeps appending to a stringbuffer which then is
+        written in the email body
+
+        Arguments:
+            record (dict): Data to send via email.
+        """
+        for _, value in record.items():
+            self.stringbuffer.append(repr(value))
+
+    def done(self):
+        """Perform final cleanup tasks.
+
+        This method is called after all records have been written. In
+        this example implementation, we properly terminate the JSON
+        array in the email body.
+
+        """
+        message = MIMEMultipart()
+        message['Date'] = email.utils.formatdate(localtime=True)
+        message['Subject'] = self.subject
+        message['From'] = self.sender
+        message['To'] = email.utils.COMMASPACE.join(self.to)
+
+        # Incase if the buffer is empty then the default message will be sent
+        if self.stringbuffer:
+            self.body = ''.join(self.stringbuffer)
+        message.attach(MIMEText(self.body))
+
+        smtp_connection = smtplib.SMTP(host=self.host, port=self.port)
+        try:
+            smtp_connection.sendmail(self.sender, self.to, message.as_string())
+        except smtplib.SMTPException as e:
+            _log.error('Failed to send email: %s', e)
+        finally:
+            smtp_connection.quit()


sunnysharmagts:
The code is ready for review and will work for both SMTP and SMTP_SSL connection.

r0hi7:
@sunnysharmagts & @susam I am just curious to know why the name `store` has been used for this plugin. What I see after getting around the code is that this will rather send the notification to the user. Is this any special case?

susam:
> @sunnysharmagts & @susam I am just curious to know why the name `store` has been used for this plugin. What I see after getting around the code is that this will rather send the notification to the user. Is this any special case?

@r0hi7 From an architecture standpoint, this belongs to the alert category, i.e., this will be specified in the `alert` part of the config. Both store and alert plugins mentioned in the config are used by `store_worker` in `workers.py`.

Do you think `EmailAlert` would be a better name for this plugin because we will use it only to send alerts about security events and we will never use it to send the original cloud records from Azure and GCP?

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,77 @@
...
+
+        """
+        message = MIMEMultipart()
+        message['Date'] = formatdate(localtime=True)
+        message['Subject'] = self.subject
+        message['From'] = self.sender
+        message['To'] = COMMASPACE.join(self.to)
+
+        #Connect to smtp host and send email
+        smtpConnection = SMTP(host=self.host, port=self.port)
```
r0hi7: For this project, we follow `snake_case`, not `camelCase`.

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,77 @@
+from email.mime.multipart import MIMEMultipart
```
r0hi7: Pylama has reported few linting issues.
```
cloudmarker/stores/emailstore.py:0:0: Incorrectly sorted imports. [isort]
cloudmarker/stores/emailstore.py:1:1: D100 Missing docstring in public module [pydocstyle]
cloudmarker/stores/emailstore.py:6:0: W0105 String statement has no effect [pylint]
cloudmarker/stores/emailstore.py:11:1: E302 expected 2 blank lines, found 1 [pycodestyle]
cloudmarker/stores/emailstore.py:11:1: D101 Missing docstring in public class [pydocstyle]
cloudmarker/stores/emailstore.py:11:0: C0111 Missing class docstring [pylint]
cloudmarker/stores/emailstore.py:13:1: D107 Missing docstring in __init__ [pydocstyle]
cloudmarker/stores/emailstore.py:13:4: R0913 Too many arguments (7/5) [pylint]
cloudmarker/stores/emailstore.py:65:9: E265 block comment should start with '# ' [pycodestyle]
cloudmarker/stores/emailstore.py:69:1: E0602 undefined name 'mailserver' [pyflakes]
cloudmarker/stores/emailstore.py:71:1: E0602 undefined name 'mailserver' [pyflakes]
cloudmarker/stores/emailstore.py:74:23: W1202 Use % formatting in logging functions and pass the % parameters as arguments [pylint]
```

You can use `make lint` command to run linter on your code. In case you want to test one specific file(e.g emailstore.py)for linting issues you can use: `pylama cloudmarker/stores/emailstore.py`.

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,77 @@
...
+    def __init__(self, host, port, subject, to, sender, body):
+        self.host = host
+        self.port = port
+        self.subject = subject
+        self.to = to
+        self.sender = sender
+        self.body = body
+
+    def write(self, record):
+        """Write JSON records to the file system.
```
r0hi7: This `write` method does nothing, but the docstring says it writes the JSON records to a file.

susam:
> Will there be a case when we will be using this plugin to send raw data from Azure or GCP, in an email?

@r0hi7 I believe there would not be a case when this plugin would send raw data from Azure or GCP to an email recipient. It would only send email alerts. So let us settle on `cloudmarker.alerts.emailalert.EmailAlert` as the class name for this plugin.

config.base.yaml
```diff
@@ -28,6 +38,9 @@ audits:
       - mockevent
     alerts:
       - filestore
+      - emailstore
+
+
```
susam: Please remove the additional blank line here.

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,77 @@
...
+        remaining records in a buffer.
+
+        """
+        message = MIMEMultipart()
+        message['Date'] = formatdate(localtime=True)
+        message['Subject'] = self.subject
+        message['From'] = self.sender
+        message['To'] = COMMASPACE.join(self.to)
+
+        #Connect to smtp host and send email
```
susam: Please remove these comments. The code you have written is already very crisp and clear and can be understood on its own, so these comments are not necessary.

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,77 @@
...
+        smtpConnection = SMTP(host=self.host, port=self.port)
+        try:
+            # identify ourselves to smtp gmail client
+            mailserver.ehlo()
```
r0hi7: `mailserver` is used without initialization.

cloudmarker/stores/emailstore.py
sunnysharmagts: Done

cloudmarker/stores/emailstore.py
sunnysharmagts: Done

cloudmarker/stores/emailstore.py
sunnysharmagts: Changed the docstring.

config.base.yaml
sunnysharmagts: Done

cloudmarker/stores/emailstore.py
sunnysharmagts: Done

cloudmarker/stores/emailstore.py
sunnysharmagts: Removed un-necessary methods.

sunnysharmagts:
> > Will there be a case when we will be using this plugin to send raw data from Azure or GCP, in an email?
> 
> @r0hi7 I believe there would not be a case when this plugin would send raw data from Azure or GCP to an email recipient. It would only send email alerts. So let us settle on `cloudmarker.alerts.emailalert.EmailAlert` as the class name for this plugin.

Let's talk about this. Just renaming the file to EmailAlert doesn't solve the whole problem. I think we should deal with alerts and persistant data storage separately. Storing anomalies in file or mongodb doesn't sound like alert to me. Rather, its kind of store. 

Now we can alert a particular user with those kind of stored data via email, stack or any other means.

susam:
> > > Will there be a case when we will be using this plugin to send raw data from Azure or GCP, in an email?
> > 
> > 
> > @r0hi7 I believe there would not be a case when this plugin would send raw data from Azure or GCP to an email recipient. It would only send email alerts. So let us settle on `cloudmarker.alerts.emailalert.EmailAlert` as the class name for this plugin.
> 
> Let's talk about this. Just renaming the file to EmailAlert doesn't solve the whole problem. I think we should deal with alerts and persistant data storage separately. Storing anomalies in file or mongodb doesn't sound like alert to me. Rather, its kind of store.
> 
> Now we can alert a particular user with those kind of stored data via email, stack or any other means.

@sunnysharmagts We can talk about this, however, we should do so in a new issue/thread. Framework design changes should be outside the scope of this pull request. In this pull request, we only care about creating an email plugin that can send the alerts, ensure that it is a minimal viable plugin that emails the alerts successfully, and then merge this pull request to master. For this pull request, I suggest picking a name that fits well within the current design and proceed.

In the current design, both store plugins and alert plugins implement the same duck-typed interface methods, `write(record)` and `done()`, and they are invoked from the `stores` and `alerts` section in the configurations. The current rationale for `stores` and `alerts` interfaces being the same are that some store plugins can also serve equally well as alert plugins. For example, the `SplunkHECStore` plugin in #47 may serve both as a store plugin and alert plugin (the alerts also need to be saved in Splunk). With the current design, there is no need to write another `SplunkHECAlert` plugin.

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,68 @@
+"""email store plugin."""
```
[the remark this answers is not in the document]
r0hi7: Also since this plugin will now be alert plugin can you please make a minor change so that path and class name looks like: `cloudmarker/alerts/emailalert.py` & `EmailAlert`

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,68 @@
...
+import logging
+from email.mime.multipart import MIMEMultipart
+from email.utils import COMMASPACE, formatdate
+from smtplib import SMTP, SMTPException
+
+# Define module-level logger.
+_log = logging.getLogger(__name__)
+
+
+class EmailStore():
```
r0hi7: This will also be changed to `EmailAlert`.

r0hi7: Since this class is not interiting some other class,  this class declaration can be made like : `class EmailAlert:` Without using `()`.

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,68 @@
...
+    """A plugin to send email notification for anomalies found."""
+
+    def __init__(self, host, port, subject, to, sender, body):
+        """Create an instance of :class:`EmailStore` plugin.
```
r0hi7: `Store` -> `Alert`

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def done(self):
+        """Perform final cleanup tasks.
+
+        This method is called after all records have been written. In
+        this example implementation, we properly terminate the JSON
+        array in the .tmp file. Then we rename the .tmp file to .json
+        file.
+
+        Note that other implementations of a store may perform tasks
```
r0hi7: Since this plugin is not implementing the behaviour of any store, what is think is doc string that starts with line 51 is not required.

config.base.yaml
```diff
@@ -17,6 +17,16 @@ events:
 alerts:
   filestore:
     plugin: cloudmarker.stores.filestore.FileStore
+  emailstore:
```
r0hi7: `store` -> `alert`

config.base.yaml
```diff
@@ -28,11 +38,11 @@ audits:
       - mockevent
     alerts:
       - filestore
+      - emailstore
```
r0hi7: `store` -> `alert`

pylama.ini
```diff
@@ -19,7 +19,11 @@ ignore = E1101
 # E1101 Instance of 'Resource' has no 'instances' member [pylint]
 
 [pylama:cloudmarker/stores/mongodbstore.py]
-ignore = R0913 
+ignore = R0913
+# R0913 Too many arguments (7/5)
+
+[pylama:cloudmarker/stores/emailstore.py]
```
r0hi7: `store` -> `alert` & `cloudmarker/stores/` ->`cloudmarker/alert/`

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,68 @@
+"""email store plugin."""
+
+
+import logging
+from email.mime.multipart import MIMEMultipart
+from email.utils import COMMASPACE, formatdate
```
r0hi7: `COMMASPACE` & `formatdate` are used only once in the code, and in this framework, we do not import functions by convention. We only import util modules and then we write `<module>.function()`.
So this line can be re-written as
```python
import email
```
And the call can be made like,
```python
message['Date'] = email.utils.formatdate(localtime=True)
```
This will also give us more clarity that `formatdata` from module `email.utils` is used here. The same can applied to `COMMASPACE` as well.

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,68 @@
...
+        array in the .tmp file. Then we rename the .tmp file to .json
+        file.
+
+        Note that other implementations of a store may perform tasks
+        like closing a connection to a remote store or flushing any
+        remaining records in a buffer.
+
+        """
+        message = MIMEMultipart()
+        message['Date'] = formatdate(localtime=True)
```
r0hi7: This will get changed to:
```python
message['Date'] = email.utils.formatdate(localtime=True)
```

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,68 @@
+"""email store plugin."""
+
+
+import logging
+from email.mime.multipart import MIMEMultipart
+from email.utils import COMMASPACE, formatdate
+from smtplib import SMTP, SMTPException
```
r0hi7: Please change this `function` level import as well as descried in previous review comment.

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,68 @@
...
+        message = MIMEMultipart()
+        message['Date'] = formatdate(localtime=True)
+        message['Subject'] = self.subject
+        message['From'] = self.sender
+        message['To'] = COMMASPACE.join(self.to)
+
+        smtp_connection = SMTP(host=self.host, port=self.port)
+        try:
+            smtp_connection.sendmail(self.sender, self.to, message.as_string())
+        except SMTPException as e:
```
r0hi7: Once ` from smtplib import SMTP, SMTPException` is changed to `import smtplib`, line number 62 and 65 will also be changed to something like : `smtplib.SMTPException`.

cloudmarker/stores/emailstore.py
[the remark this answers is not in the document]
susam: Please also write the module level docstrings with correct capitalization. Please see `MongoDBStore`, `FileStore`, etc. as examples. These docstrings are used to generate project documentation that gets published automatically at https://cloudmarker.readthedocs.io/ , so it is all the more important that these docstrings use correct capitalization in particular and look good in general.

cloudmarker/stores/emailstore.py
```diff
@@ -0,0 +1,68 @@
...
+        self.sender = sender
+        self.body = body
+
+    def write(self, record):
+        """Add record to a buffer to sent via email.
+
+        This method does nothing for now. However, this method
+        show keep adding all the records to the buffer to
+        send via email to particular recipient.
+        """
```
susam: The implementation of the `write()` method needs to be completed such that the `done()` method can send emails successfully in order for this pull request to be a minimum viable plugin. We will be merging this pull request after this pull request is functional and this plugin can send emails.

cloudmarker/stores/emailstore.py
sunnysharmagts: Done

sunnysharmagts: Done

cloudmarker/stores/emailstore.py
sunnysharmagts: Done

cloudmarker/stores/emailstore.py
sunnysharmagts: As per other comments given by @susam , I have implemented the write method and provided with proper docstring

config.base.yaml
sunnysharmagts: Done

config.base.yaml
sunnysharmagts: Done

cloudmarker/stores/emailstore.py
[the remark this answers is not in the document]
sunnysharmagts: > Also since this plugin will now be alert plugin can you please make a minor change so that path and class name looks like: `cloudmarker/alerts/emailalert.py` & `EmailAlert`

I am a little confused here.  As per your comment `cloudmarker/alerts/emailalert.py`, do you want the alerts to be in the different alert package than the store package ?

cloudmarker/stores/emailstore.py
sunnysharmagts: At some level it makes sense to follow this approach. But I googled about both the approaches and found out that there are respective pros and cons of using the respective import convention.

But then here you are trying to import full `email` package instead of just import utils. The main question is the use of any particular lib which you are trying to import. It doesn't make sense to import full email package when I am just using COMMASPACE, formatedate from the utils packages. All the modules inside email is getting referenced which are not used at all.

So I believe, we should only import specific lib which we are using.

cloudmarker/stores/emailstore.py
sunnysharmagts: Currently, I am changing this as asked to, but I disagree which the approach of importing full library. We can have a discussion on this.

cloudmarker/stores/emailstore.py
sunnysharmagts: Same as comments above.

cloudmarker/stores/emailstore.py
sunnysharmagts: Same as comments above.

cloudmarker/stores/emailstore.py
susam: > It doesn't make sense to import full email package when I am just using COMMASPACE, formatedate from the utils packages.

Why does it not make sense? Please provide some justification.

> All the modules inside email is getting referenced which are not used at all.

We have to import `email.utils` (not the `email` package). If we want to use something like `COMMASPACE` or `formatdate`, the Python language requires that we import the `email.utils` module. Importing the `email` package won't make `COMMASPACE` or `formatdate` available.

So if we want to use something like `COMMASPACE` or `formatdate`, there are three popular options:

1. `from email.utils import COMMASPACE, formatdate` followed by using `COMMASPACE` an `formatdate` later in the code.
2. `import email.utils` followed by using `email.utils.COMMASPACE` and `email.utils.formatdate`.
3. `from email import utils` followed by using `utils.COMMASPACE` and `utils.formatdate`

Option 1 violates our coding convention. We avoid option 1. The reason for avoiding option 1 is the benefits offered by options 2 and 3 discussed below.

Option 2 can sometimes be too verbose because we have to write the fully qualified indentifier every time we use it.

Option 3 is preferred usually. When we see something like `utils.formatdate` in the code, it tells us clearly that `formatdate` is not written in the current module but instead imported from another module. This kind of contextual  information is helpful in reducing cognitive load while debugging a complex issue. For example, we can quickly eliminate `formatdate` as a likely suspect for a bug because we know it has come from a well tested standard library.

Also, note that in all three cases, the entire `email.utils` package is imported anyway.

Having said that in this specific case, I would choose Option 2 even if it is a little too verbose. The reason is that `utils` is such a common name that `utils.formatdate` still keeps us confused whether it is our `utils` module or a standard/thirdparty library's `utils` module. On the other hand, `email.utils.formatdate` is very explicit and tells us accurately where `formatdate` is defined.

cloudmarker/stores/emailstore.py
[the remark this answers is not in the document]
susam: If a module is going to be used an alert plugin, I think it makes sense to have it in `cloudmarker.alerts` package.

cloudmarker/stores/emailstore.py
sunnysharmagts: Got the concept of importing functions and classes. So whatever you mentioned now makes sense.

cloudmarker/stores/emailstore.py
sunnysharmagts: Added the implementation in the `write()` method and also wrote proper doc string for that.

pylama.ini
sunnysharmagts: A little confusion here. Do you want alerts to be in the separate package ?

cloudmarker/stores/emailstore.py
susam: I haven't understood the disagreement entirely. If we want to use an identifier from another module, we have to import the full module anyway. There is no way to avoid that. That's just how Python is designed.

It appears to me that the disagreement is not regarding whether to import a full module or not but about the coding style on whether to write fully qualified identifiers or unqualified identifiers.

cloudmarker/stores/emailalert.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def write(self, record):
+        """Write JSON records to the file system.
+
+        This method is called once for every ``record`` read from a
+        cloud. In this example implementation of a store, we simply
+        send the ``record`` in JSON format via email to the recipient.
+        The records keeps appending to a stringbuffer which then is
+        written in the email body
+        Arguments:
```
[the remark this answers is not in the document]
susam: Note that if we run `make docs` it would automatically modify `docs/api/cloudmarker.stores.rst` to automatically include the documentation of this plugin in the generated documentation. It's a good idea to add this file and commit it too as part of this pull request.

susam:
The commit message still does not conform to the conventions discussed in [this comment](https://github.com/cloudmarker/cloudmarker/pull/55#pullrequestreview-218139728). Especially, note this:

> If there is only one point to discuss in the commit message, then there is no need of a bullet point. The whole point can be written as a single paragraph.

The commit message needs to be updated:

```
git commit --amend
```

If more pull requests get merged into `master` before this pull request is merged, then this pull request also needs to be rebased on the latest `master`.

```
git checkout master
git pull upstream master
git push origin master
git checkout notification
git rebase master
```

In future, please consider following the branch-and-pull-request workflow, i.e., clone `cloudmarker/cloudmarker`, create a branch in it, then push it to `origin`, and create a pull request. With the branch-and-pull-request workflow, I can perform these routine Git tasks on your behalf and I would not have to bother you with these requests. With the fork-and-pull-request workflow, you would have to do this amending and rebasing work for us.

docs/api/cloudmarker.stores.rst
```diff
@@ -9,6 +9,14 @@ cloudmarker.stores package
 Submodules
 ----------
 
+cloudmarker.stores.emailalert module
```
r0hi7: I think this needs to be corrected.

susam: Yes, there is issue #72 open to track this.

PR #59, opened 2019-03-26
SlackAlert Plugin

r0hi7:


susam:
Rebased this pull request on the latest `master`.

cloudmarker/alerts/slackalert.py
```diff
@@ -0,0 +1,108 @@
...
+class SlackAlert:
+    """Alert plugin to send Slack notifications."""
+
+    def __init__(self, bot_user_token, to, text,
+                 temp_file='/tmp/cloudmarker/slackanomalyreport.json'):
+        """Initialize the class:`SlackAlerter`.
+
+        Arguments:
+          bot_user_token (string): Token for Slack bot user.
+          to (list): List of recipients(string) to send Slack alert to.
```
susam: Please add a space between "recipients" and "(string)".

cloudmarker/alerts/slackalert.py
```diff
@@ -0,0 +1,108 @@
...
+import slackclient
+
+_log = logging.getLogger(__name__)
+
+
+class SlackAlert:
+    """Alert plugin to send Slack notifications."""
+
+    def __init__(self, bot_user_token, to, text,
+                 temp_file='/tmp/cloudmarker/slackanomalyreport.json'):
```
susam: Please rename the file path to `/tmp/cloudmarker/slackalert.json` to keep it simple and consistent with the plugin name.

cloudmarker/alerts/slackalert.py
```diff
@@ -0,0 +1,108 @@
...
+                                         user=user_id)['channel']['id']
+
+                sc.api_call(
+                    'chat.postMessage',
+                    channel=channel_id,
+                    attachments=[
+                        {
+                            'fallback': 'Cloudmarker Notification!',
+                            'color': '#36a64f',
+                            'title': 'Cloudmarker Anomaly Notification'
```
susam: A couple of comments here:

- Let the title just be `Cloudmarker Alert`.
- Search for "notification" or "Notification" in the entire file and change it to "alert" or "Alert". The reason is that we are settling on the term "alert" and we will use it consistently everywhere.
- Please add a comment above that explains how the `fallback` and `title` options are used or what they mean.

cloudmarker/alerts/slackalert.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __init__(self, bot_user_token, to, text,
+                 temp_file='/tmp/cloudmarker/slackanomalyreport.json'):
+        """Initialize the class:`SlackAlerter`.
+
+        Arguments:
+          bot_user_token (string): Token for Slack bot user.
+          to (list): List of recipients(string) to send Slack alert to.
+          text (string): Message body.
+          temp_file (string): Name of file to be used to save intrim JSON
```
susam: `intrim` -> `interim`.

config.base.yaml
```diff
@@ -28,6 +35,7 @@ audits:
       - mockevent
     alerts:
       - filestore
+      - slackalert
```
susam: Please remove `slackalert` from the audit section. The `slackalert` definition at line 20 above should remain. Only this one at line 38 should be removed. The reason is that after we clone `cloudmarker` from GitHub repo, it should run without any errors without any further configuration. Since `slackalert` requires additional configuration, it should be added by the user only if they want to do so.

PR #64, opened 2019-03-28
Rename sub_id to subscription_id in AzureCloud

### LINKED ISSUE #60: Rename sub_id to subscription_id in AzureCloud
The key `sub_id` seems too short and may not be obvious to what "sub" expands to. I think it is good to name it `subscription_id` instead as per the Azure terminology.

r0hi7:
The `sub_id` field has been renamed to subscription_id in AzureCloud plugin.
Resolves #60


susam:
Rebased on the latest `master`. Also, updated the commit message to `Rename sub_id to subscription_id in AzureCloud`.

PR #65, opened 2019-03-29
Generate firewall rules with common properties

susam:
Network security group (NSG) records from Azure and firewall records
from GCP are in very different formats. In both types of records, each
record may contain multiple firewall rules. This change ensures that
both AzureCloud and GCPCloud plugin generates a new record for each
firewall rule found in NSG/firewall record.

Further, this change ensures that the `com` bucket, i.e.,
`record['com']` is populated with firewall rule properties that we care
about in common notation, where `record` denotes a firewall rule record.

### DIFF
--- a/cloudmarker/clouds/gcpcloud.py
+++ b/cloudmarker/clouds/gcpcloud.py
@@ -6,6 +6,7 @@ Google Cloud Platform.
 
 
 import json
+import logging
 
 from google.oauth2 import service_account
 from googleapiclient import discovery
@@ -15,6 +16,9 @@ _GCP_SCOPES = ['https://www.googleapis.com/auth/compute.readonly']
 
 See https://developers.google.com/identity/protocols/googlescopes for
 more details on OAuth 2.0 scopes for Google APIs."""
+
+
+_log = logging.getLogger(__name__)
 
 
 class GCPCloud:
@@ -71,7 +75,7 @@ class GCPCloud:
                                credentials=self._credentials,
                                cache_discovery=False)
 
-    def _get_firewall_rules(self):
+    def _get_firewalls(self):
         """Query the compute resource and return firewall rules.
 
         Returns:
@@ -150,61 +154,23 @@ class GCPCloud:
         Yields:
             dict: Firewall rule or VM instance configuration data.
 
-        Here is an example of a firewall rule record :obj:`dict` yielded
-        by this method:
-
-        .. code:: python
-
-            {
-              "kind": "compute#firewall",
-              "id": "7890789078907890",
-              "creationTimestamp": "2018-12-19T01:43:51.988-08:00",
-              "name": "default-allow-ss",
-              "description": "str",
-              "network": "https://www.googleapis.com/compute/v1/"
-                         "projects/foo/global/networks/default",
-              "priority": 1000,
-              "sourceRanges": [
-                "0.0.0.0/0"
-              ],
-              "targetTags": [
-                "https-server"
-              ],
-              "allowed": [
-                {
-                  "IPProtocol": "tcp",
-                  "ports": [
-                    "80",
-                    "8080-8090"
-                  ]
-                }
-              ],
-              "direction": "INGRESS",
-              "logConfig": {
-                "enable": bool
-              },
-              "disabled": bool,
-              "selfLink": "https://www.googleapis.com/compute/v1/"
-                          "projects/foo/global/firewalls/default-allow-https",
-              "record_type": "firewall_rule"
-            }
-
         """
-        firewall_rules = self._get_firewall_rules()
+        firewalls = self._get_firewalls()
         instances = self._get_instances()
 
-        for rule in firewall_rules:
+        for firewall in firewalls:
             record = {
-                'raw': rule,
+                'raw': firewall,
                 'ext': {
-                    'record_type': 'firewall_rule'
+                    'record_type': 'firewall'
                 },
                 'com': {
                     'cloud_type': 'gcp',
-                    'record_type': 'firewall_rule'
+                    'record_type': None,
                 }
             }
             yield record
+            yield from _get_normalized_firewall_rules(firewall)
 
         for instance in instances:
             record = {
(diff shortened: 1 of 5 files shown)


cloudmarker/clouds/gcpcloud.py
```diff
@@ -150,61 +154,23 @@ def read(self):
...
-                    'record_type': 'firewall_rule'
+                    'record_type': 'firewall'
                 },
                 'com': {
                     'cloud_type': 'gcp',
-                    'record_type': 'firewall_rule'
+                    'record_type': None,
                 }
             }
             yield record
```
r0hi7: Why are you yielding the original record, even after the record is split in multiple records after normalization?

cloudmarker/clouds/gcpcloud.py
```diff
@@ -150,61 +154,23 @@ def read(self):
...
-                'raw': rule,
+                'raw': firewall,
                 'ext': {
-                    'record_type': 'firewall_rule'
+                    'record_type': 'firewall'
                 },
                 'com': {
                     'cloud_type': 'gcp',
-                    'record_type': 'firewall_rule'
+                    'record_type': None,
```
r0hi7: Why you have changed the record_type to `null` here?

cloudmarker/clouds/gcpcloud.py
```diff
@@ -226,3 +192,112 @@ def done(self):
...
+        }
+    }
+    return record
+
+
+def _get_normalized_firewall_direction(firewall):
+    rule_name = firewall.get('name')
+    direction = firewall.get('direction')
+
+    if direction is None:
```
r0hi7: Will there be any case when the direction of firewall rule in either of Azure or GCP cloud will be `None`?

cloudmarker/clouds/gcpcloud.py
```diff
@@ -226,3 +192,112 @@ def done(self):
...
+                _get_normalized_firewall_direction(firewall),
+
+            'protocol':
+                _get_normalized_firewall_protocol(rule),
+
+            'source_addresses': firewall.get('sourceRanges'),
+
+            # If the 'ports' key is missing in an allowed/denied rule,
+            # it means all ports are allowed/denied.
+            'destination_ports': rule.get('ports', ['0-65535'])
```
r0hi7: How about using `*` in place of `0-65535`?

cloudmarker/clouds/gcpcloud.py
```diff
@@ -226,3 +192,112 @@ def done(self):
...
+        for rule in allow_rules:
+            firewall_rule = _get_normalized_firewall_rule(firewall, rule)
+            firewall_rule['com']['access'] = 'allow'
+            yield firewall_rule
+
+    deny_rules = firewall.get('denied')
+    if deny_rules is not None:
+        for rule in deny_rules:
+            firewall_rule = _get_normalized_firewall_rule(firewall, rule)
+            firewall_rule['com']['access'] = 'allow'
```
r0hi7: Why the `access` for `deny_rules` is also marked as `allow`?

cloudmarker/test/test_gcpcloud.py
```diff
@@ -112,3 +112,287 @@ def test_instance_with_next_page_token(
         list(gcpcloud.GCPCloud('', '').read())
 
         self.assertEqual(mock_execute.mock_calls, [mock.call(), mock.call()])
+
+    def _patch(self, target):
+        patcher = mock.patch('cloudmarker.clouds.gcpcloud.' + target)
+        self.addCleanup(patcher.stop)
+        return patcher.start()
+
+    def setUp(self):
```
r0hi7: I really like this approach to patching the test_methods. I was completely unaware of this. Thanks for introducing this to me :)

cloudmarker/clouds/gcpcloud.py
susam: The original record is yielded here so that it remains preserved in our data store. We do not have storage constraint right now. Most of the data stores supported by this project would be big data systems. Also, the original raw `firewall` record has more information than what these normalized `firewall_rule` records have. So I do not see a good reason to lose the original `firewall` data.

Also, here at [line 252](https://github.com/cloudmarker/cloudmarker/pull/65/files#diff-c403e664c38d6edc528237cc5e434e05R252), we are keeping a `reference` to the original raw record in the `firewall_rule` record. In the cloud-independent `FirewallRuleEvent` plugin in #69, this `reference` would be included in every event generated. Thus, when a user of this project (a security analyst for example) reads an alert, he/she can find the original raw `firewall` record using this `reference` if he/she wishes too.

cloudmarker/clouds/gcpcloud.py
susam: The properties in the `com` bucket needs to be well thought out. The `com` bucket is something like what CIM of Splunk achieves. It defines standardized property names that would work for all clouds.

In Azure, we have an `nsg` record that has a `security_rules` list with multiple security rules. We take an `nsg` record and split each security rule into its own `firewall_rule` record.

In GCP, we have a `firewall` record (represented as `"kind": "compute#firewall"` in the JSON). This `firewall` record has an `allowed` or `denied` list with multiple port rules. We take a `compute#firewall` record and split each port rule into its own `firewall_rule` record.

In case of `firewall_rule` record, we have complete control on each property in it because we are generating it ourselves. So we are populating its `com` bucket with well thought out names.

But what do we do for the original `nsg` and `compute#firewall` records? They are incompatible with each other. What common `record_type` can we enter that holds good for both? Particularly,

- Should we say `record['com']['record_type']` is `nsg` for Azure but `firewall` for GCP? In that case, someone searching for `record['com']['record_type'] == 'nsg'` would get the Azure NSGs but not the GCP firewall records. Is this good behavior? I don't know.
- Or should we we say `record['com']['record_type']` is `firewall` for both Azure and GCP? In that case a search for `record['com']['record_type'] == 'firewall'` would return both Azure NSGs and GCP firewalls but then there is no other meaningful information about the firewall in the `com` bucket, so what purpose does setting `record['com']['record_type']` to `firewall` for both serve?

In a nutshell, I could not arrive at any good decision for `record['com']['record_type']` value for Azure NSG and GCP firewall. So rather than putting something that may not make sense in one situation or another, I have left it as `None` for now. We can always set it to something useful later provided we can come up with a rationale for it that is consistent for both Azure and GCP.

On the other hand, the `ext` bucket contains record types that makes sense for both Azure and GCP. For example, `record['ext']['record_type']` is `security_rule` for Azure and `firewall` for GCP. The `ext` bucket (extended bucket) is meant for keeping cloud-specific property names and values like this.

The `com` bucket on the other hand is meant for keeping standardized property names and values that work uniformly for all clouds.

cloudmarker/clouds/gcpcloud.py
susam: There should not be any case where `direction` is `None`. However, the data returned to us by Azure is not in our control. When we are dealing with data that is not in our control, it's a good practice to write our code as defensively as possible with as little assumptions as possible.

In the above code, we do not make any assumptions about whether `direction` key is present or not in the response from GCP Cloud. With only about three lines of overhead, we are able to gracefully handle a scenario of missing `direction` key. So I think these three lines of overhead are worth it. It avoids the possibility of the service crashing unexpectedly if some day for some reason the `direction` key goes missing in the GCP response.

cloudmarker/clouds/gcpcloud.py
susam: Here's my rationale for choosing `0-65535` instead of `*`.

- In the standardized notation in `com` bucket, we want to simplify the notation, so if we can represent the notion of all ports with `0-65535`, there is no strict necessity of introducing a new notation like `*` to represent the same concept.
- A security analyst as well as event plugins would rely on the property names and values in `com` bucket to search and detect security issues. If we populate this `destination_ports` field with two different types of notation (an `M-N` notation as well as `*` notation), then the security analysts have to learn and use both notations while searching the `com` bucket and similarly our event plugin also needs to be able to support both types of notation. Choosing a single notation simplifies the work for both security analysts and event plugins.
- We already have two types of notation already in `destination_ports`: `M` notation (e.g, `80`) and `M-N` (e.g., `8000-8080`). These two notations are imposed on us by the existing data format in Azure on GCP. One can argue that we could have got rid of the `M` notation by writing `M` as `M-M` (i.e., writing `80` as `80-80` instead) but that is not a convenient or intuitive notation to work with for a human security analyst. So I decided to keep two separate notations: `M` notation for single port and `M-N` notation for port range. Since the notion of all ports is taken care of by `M-N` notation which is intuitive enough for the human user, I decided not to introduce a `*` notation.
- There is no specific benefit I see for supporting the `*` notation but there is at least one drawback (2nd point above).

cloudmarker/clouds/gcpcloud.py
susam: This is a mistake. Thanks for noticing this. I will update the pull request and fix it.

cloudmarker/test/test_gcpcloud.py
susam: Yes, this helps in reducing the boilerplate required for patching the same objects for every test method and specifying the same mock arguments for every test method.

By the way, patching the same objects for every test method can be avoided by specifying the patch decorator on top of the entire test class. But we still need to specify the same mock arguments for every test method.

This approach takes care of avoiding both types of boilerplate. It's good to know that you like it.

cloudmarker/clouds/gcpcloud.py
susam: Fixed it at [line 230](https://github.com/cloudmarker/cloudmarker/pull/65/files#diff-c403e664c38d6edc528237cc5e434e05R230).

Also added an assert for it in the tests at [line 208](https://github.com/cloudmarker/cloudmarker/pull/65/files#diff-831ed9aecd6b6e0d34f3fb816425070bR208).

PR #66, opened 2019-03-30
Rename _logger to _log in workers

### LINKED ISSUE #61: Rename _logger to _log in workers
The module-level logger variable is named as `_log` in most places but `_logger` in one place.

```
$ grep -r "logging.getLogger" cloudmarker
cloudmarker/stores/mongodbstore.py:_log = logging.getLogger(__name__)
cloudmarker/clouds/azurecloud.py:_log = logging.getLogger(__name__)
cloudmarker/workers.py:_logger = logging.getLogger(__name__)
cloudmarker/manager.py:_log = logging.getLogger(__name__)
```

Since `_log` is the more frequently used name, we should settle on using `_log` everywhere.

Rename `_logger` to `_log` in `workers.py`.

r0hi7:
Resolves #61

### DIFF
--- a/cloudmarker/workers.py
+++ b/cloudmarker/workers.py
@@ -12,7 +12,7 @@ queues necessary to pass records from one plugin class to another.
 
 import logging
 
-_logger = logging.getLogger(__name__)
+_log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


susam:
Rebased on the latest `master`.

PR #68, opened 2019-03-31
Add logs in GCPCloud

### LINKED ISSUE #63: Add logs in GCPCloud
See the "Found" logs in AzureCloud plugin. GCPCloud plugin should also have similar logs.

r0hi7:
Resolves #63

### DIFF
--- a/cloudmarker/clouds/gcpcloud.py
+++ b/cloudmarker/clouds/gcpcloud.py
@@ -158,6 +158,11 @@ class GCPCloud:
         firewalls = self._get_firewalls()
         instances = self._get_instances()
 
+        _log.info('Found %d firewall records for project %s',
+                  len(firewalls), self._project_name)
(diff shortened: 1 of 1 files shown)


cloudmarker/clouds/gcpcloud.py
```diff
@@ -6,10 +6,13 @@
 
 
 import json
+import logging
 
 from google.oauth2 import service_account
 from googleapiclient import discovery
 
+_log = logging.getlogger(__name__)
```
susam: Does this code work for you? This line should lead to the following error:

```
AttributeError: module 'logging' has no attribute 'getlogger'
```

The actual function name is `getLogger`.

r0hi7: I did not run the code after making changes :P

r0hi7:
@susam I have updated, please review now.

Issue #53, opened 2019-03-17
Send email notification after all audits have run


susam:
Send an email notification after all audits have run.

susam:
Changed the assignee of this issue to myself after talking to @sunnysharmagts.  Thanks @sunnysharmagts.

I would like to make some progress on this as soon as possible because this is a necessary framework feature we are going to need in the next release. Completing this work soon would also help us to finalize the feature set for a new release, so that I can document the feature set in pull request #79.

PR #75, opened 2019-04-10
Add SSL and authentication support in EmailAlert

sunnysharmagts:
Add smtpssl support for sending email with authentication via gmail.

### DIFF
--- a/cloudmarker/alerts/emailalert.py
+++ b/cloudmarker/alerts/emailalert.py
@@ -14,25 +14,31 @@ _log = logging.getLogger(__name__)
 class EmailAlert:
     """A plugin to send email notification for anomalies found."""
 
-    def __init__(self, host, port, subject, to, sender, body):
+    def __init__(self, host, port, subject, to, sender, body, use_ssl=True,
+                 username=None, password=None):
         """Create an instance of :class:`EmailAlert` plugin.
 
         Arguments:
-            host (string): Server hostname for SMTP.
+            host (str): Server hostname for SMTP.
             port (int): Server port for SMTP.
-            subject (string): Subject line for email.
-            to (list): List of recipients (string).
-            sender (string): From field for email.
-            body (string): Email body.
-
+            subject (str): Subject line for email.
+            to (list): List of recipients (str).
+            sender (str): From field for email.
+            body (str): Email body.
+            use_ssl (bool): Is SSL connection required
+            username (str): username of email account
+            password (str): password of email account
         """
-        self.host = host
-        self.port = port
-        self.subject = subject
-        self.to = to
-        self.sender = sender
-        self.body = body
-        self.stringbuffer = []
+        self._host = host
+        self._port = port
+        self._subject = subject
+        self._to = to
+        self._sender = sender
+        self._body = body
+        self._stringbuffer = []
+        self._use_ssl = use_ssl
+        self._username = username
+        self._password = password
 
     def write(self, record):
         """Write JSON records to the file system.
@@ -47,7 +53,7 @@ class EmailAlert:
             record (dict): Data to send via email.
         """
         for _, value in record.items():
-            self.stringbuffer.append(repr(value))
+            self._stringbuffer.append(repr(value))
 
     def done(self):
         """Perform final cleanup tasks.
@@ -56,22 +62,58 @@ class EmailAlert:
         this example implementation, we properly terminate the JSON
         array in the email body.
 
+        The connection is made based whether the SMTP or SMTP_SSL is
+        required which is determined by ``use_ssl`` param. The SMTP
+        login process is conducted if ``username`` param is provided.
+
         """
         message = MIMEMultipart()
         message['Date'] = email.utils.formatdate(localtime=True)
-        message['Subject'] = self.subject
-        message['From'] = self.sender
-        message['To'] = email.utils.COMMASPACE.join(self.to)
+        message['Subject'] = self._subject
+        message['From'] = self._sender
+        message['To'] = email.utils.COMMASPACE.join(self._to)
 
         # Incase if the buffer is empty then the default message will be sent
-        if self.stringbuffer:
-            self.body = ''.join(self.stringbuffer)
-        message.attach(MIMEText(self.body))
+        if self._stringbuffer:
+            self._body = ''.join(self._stringbuffer)
+        message.attach(MIMEText(self._body))
 
-        smtp_connection = smtplib.SMTP(host=self.host, port=self.port)
+        smtp = self._prepare_smtp_session()
         try:
-            smtp_connection.sendmail(self.sender, self.to, message.as_string())
+            smtp.sendmail(self._sender, self._to, message.as_string())
         except smtplib.SMTPException as e:
             _log.error('Failed to send email: %s', e)
         finally:
-            smtp_connection.quit()
+            smtp.quit()
+
+    def _prepare_smtp_session(self):
+        """Return SMTP connection object.
+
+        Create a SMTP connection based on whether the SSL param.
+        If SSL connection is required (for eg: in case of gmail) then
+        method will return a SMTP_SSL connection. In other cases if the
+        authentication is not required or SMTP_SSL connection is not
+        required then a plain SMTP connection object is returned.
+        For gmail follow the steps given below even to work for 2 factor
+        authentication
+        1. Log-in into Gmail with your account
+        2. Navigate to https://security.google.com/settings/security/
+        apppasswords
+        3. In 'select app' choose 'custom', give it an arbitrary name and
+        press generate
+        4. It will give you 16 chars token.
+        5. Use that token as password for login
+        6. Host is smtp.gmail.com and port is 465
+
+        """
+        if self._use_ssl:
+            smtp = smtplib.SMTP_SSL(host=self._host, port=self._port)
+            smtp.set_debuglevel(True)
+        else:
+            smtp = smtplib.SMTP(host=self._host, port=self._port)
+            smtp.set_debuglevel(True)
+            smtp.starttls()
+
+        if self._username:
+            smtp.login(self._username, self._password)
+        return smtp
--- a/cloudmarker/baseconfig.py
+++ b/cloudmarker/baseconfig.py
@@ -53,6 +53,19 @@ events:
 alerts:
   filestore:
     plugin: cloudmarker.stores.filestore.FileStore
+  emailalert:
+    plugin: cloudmarker.alerts.emailalert.EmailAlert
+    params:
+      use_ssl: True
+      host: smtp.example.com
+      port: 25
+      sender: CloudMarker Alert <no-reply@example.com>
+      to:
+        - recipient@example.com
+      subject: Anomaly notification
+      body: default text
+      username: alice
+      password: alice@123
 
 audits:
   mockaudit:
--- a/pylama.ini
+++ b/pylama.ini
@@ -28,7 +28,9 @@ ignore = R0911
 # R0911 Too many return statements (8/6) [pylint]
 
 [pylama:cloudmarker/alerts/emailalert.py]
-ignore = R0913,C0412
+ignore = R0913,C0412,R0902
+
+# R0902 Too many instance attributes (10/7) [pylint]
 
 [pylama:cloudmarker/stores/splunkhecstore.py]
 ignore = R0913,W0703


cloudmarker/alerts/emailalert.py
```diff
@@ -68,10 +79,53 @@ def done(self):
             self.body = ''.join(self.stringbuffer)
         message.attach(MIMEText(self.body))
 
-        smtp_connection = smtplib.SMTP(host=self.host, port=self.port)
+        smtp_connection = self.get_connection()
```
susam: The variable name `smtp_connection` is slightly non-idiomatic while working with `smtplib` or sockets/connections in general. A more idiomatic name would be just `smtp` or `server`. I suggest renaming this to `smtp`.

Since this variable lives in a very small scope, I think we can make this variable name short and crisp.

cloudmarker/alerts/emailalert.py
```diff
@@ -68,10 +79,53 @@ def done(self):
...
+            # an ehlo call is required just after the connect method and just
+            # before login
+            # just to identity yourself. If not called it would throw a
+            # SMTPSenderRefused error
+            # Failed to send email; error: SMTPSenderRefused: (503, b'5.5.1
+            # EHLO/HELO first.
+            smtp_connection.ehlo()
+            smtp_connection.login(self.username, self.password)
+            return smtp_connection
+        return smtplib.SMTP(host=self.host, port=self.port)
```
susam: If the code is simplified as described in the previous comments, then this whole separate function would no longer be required. It would be a simple 4-line if-then code which can just be moved back to the `write()` method.

config.base.yaml
```diff
@@ -27,13 +27,16 @@ alerts:
...
+      is_ssl_required: False
       host: smtp.example.com
       port: 25
       sender: CloudMarker Alert <no-reply@example.com>
       to:
         - recipient@example.com
       subject: Anomaly notification
       body: default text
+      username: username
+      password: password
```
susam: Let us get a little bit creative here to very clearly highlight the strings that need to be changed. My suggestions:

```
username: alice
password: hunter2
```

docs/api/cloudmarker.stores.rst
```diff
@@ -9,14 +9,6 @@ cloudmarker.stores package
...
-cloudmarker.stores.emailalert module
-------------------------------------
-
-.. automodule:: cloudmarker.stores.emailalert
-    :members:
-    :undoc-members:
-    :show-inheritance:
-
 cloudmarker.stores.filestore module
 -----------------------------------
```
susam: The `emailalert` documentation should appear in `cloudmarker.alerts.rst`, so if this deletion change is included in this PR, the corresponding addition change should also be included in this PR.

If you just run `make docs`, that should do the necessary updates to make these changes ready.

pylama.ini
```diff
@@ -28,7 +28,7 @@ ignore = R0911
 # R0911 Too many return statements (8/6) [pylint]
 
 [pylama:cloudmarker/alerts/emailalert.py]
-ignore = R0913,C0412
+ignore = R0913,C0412,R0902
```
susam: What do these ignore? Please add comments below this `ignore` line for each warning being ignored. See the rest of the file for these commenting conventions.

cloudmarker/alerts/emailalert.py
```diff
@@ -14,7 +14,8 @@
 class EmailAlert:
     """A plugin to send email notification for anomalies found."""
```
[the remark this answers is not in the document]
susam: Note: I fixed the author/committer details myself yesterday and force-pushed the new commit, so this review comment is taken care of. Other review comments still need to be addressed.

cloudmarker/alerts/emailalert.py
```diff
@@ -68,10 +79,53 @@ def done(self):
...
-        smtp_connection = smtplib.SMTP(host=self.host, port=self.port)
+        smtp_connection = self.get_connection()
         try:
             smtp_connection.sendmail(self.sender, self.to, message.as_string())
         except smtplib.SMTPException as e:
             _log.error('Failed to send email: %s', e)
         finally:
             smtp_connection.quit()
+
+    def get_connection(self):
```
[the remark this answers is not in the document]
sunnysharmagts: Done in commit https://github.com/cloudmarker/cloudmarker/commit/1f9bdb2a7ca47df4c78073bc84f621ee29017f04#diff-28d4faa43f2d620f5a2c54123370ee47R82

cloudmarker/alerts/emailalert.py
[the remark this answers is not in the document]
sunnysharmagts: Done https://github.com/cloudmarker/cloudmarker/commit/11ff34083ba27900832dc7daf05a72cd2a99caca

cloudmarker/alerts/emailalert.py
```diff
@@ -24,7 +25,9 @@ def __init__(self, host, port, subject, to, sender, body):
             to (list): List of recipients (string).
             sender (string): From field for email.
             body (string): Email body.
-
+            is_ssl_required (bool): Is SSL Connection required
+            username (string): username of email account
```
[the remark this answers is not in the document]
sunnysharmagts: Done https://github.com/cloudmarker/cloudmarker/commit/11ff34083ba27900832dc7daf05a72cd2a99caca#diff-28d4faa43f2d620f5a2c54123370ee47R24

cloudmarker/alerts/emailalert.py
```diff
@@ -24,7 +25,9 @@ def __init__(self, host, port, subject, to, sender, body):
...
+            password (password): password of email account
```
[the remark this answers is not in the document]
sunnysharmagts: Done https://github.com/cloudmarker/cloudmarker/commit/11ff34083ba27900832dc7daf05a72cd2a99caca#diff-28d4faa43f2d620f5a2c54123370ee47R30

cloudmarker/alerts/emailalert.py
```diff
@@ -14,7 +14,8 @@
 class EmailAlert:
     """A plugin to send email notification for anomalies found."""
 
-    def __init__(self, host, port, subject, to, sender, body):
+    def __init__(self, host, port, subject, to, sender, body,
+                 is_ssl_required, username=None, password=None):
```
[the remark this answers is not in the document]
sunnysharmagts: Done https://github.com/cloudmarker/cloudmarker/commit/11ff34083ba27900832dc7daf05a72cd2a99caca#diff-28d4faa43f2d620f5a2c54123370ee47R39

cloudmarker/alerts/emailalert.py
```diff
@@ -68,10 +79,53 @@ def done(self):
...
+        if self.is_ssl_required:
+
+            smtp_connection = smtplib.SMTP_SSL(host=self.host, port=self.port)
+            smtp_connection.set_debuglevel(True)
+
+            # identity yourself to esmtp server using ehlo. This call is
+            # required
+            # or else it would throw error: SMTPSenderRefused: (503, b'5.5.1
+            # EHLO/HELO first.
+            smtp_connection.ehlo()
```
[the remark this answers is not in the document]
sunnysharmagts: Done. Removed the ehlo() method call

cloudmarker/alerts/emailalert.py
```diff
@@ -68,10 +79,53 @@ def done(self):
...
+            smtp_connection.connect(self.host, self.port)
```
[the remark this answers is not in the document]
sunnysharmagts: Done

cloudmarker/alerts/emailalert.py
sunnysharmagts: https://github.com/cloudmarker/cloudmarker/commit/11ff34083ba27900832dc7daf05a72cd2a99caca#diff-28d4faa43f2d620f5a2c54123370ee47R112

cloudmarker/alerts/emailalert.py
sunnysharmagts: I have removed the `ehlo` methods but I am keeping the `_get_connection` method. The reason for doing so is that it creates the connection based on the `use_ssl` params and returns the object. Also it contains the comments to do so. 

So I believe its better to keep this as a separate method rather then flooding the `write` method.

config.base.yaml
sunnysharmagts: https://github.com/cloudmarker/cloudmarker/commit/11ff34083ba27900832dc7daf05a72cd2a99caca#diff-abc06351b88de7752624acd35f48332cR40

pylama.ini
sunnysharmagts: https://github.com/cloudmarker/cloudmarker/commit/11ff34083ba27900832dc7daf05a72cd2a99caca#diff-21c09aff342003d14df85a14eeaae393R33

cloudmarker/alerts/emailalert.py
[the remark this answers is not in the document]
susam: @sunnysharmagts This does not look right. The commit message is `SSL support SMTP in EmailAlert` which does not look like a proper statement. The commit message needs to be imperative mood. See https://github.com/erlang/otp/wiki/writing-good-commit-messages for commit message conventions.

cloudmarker/alerts/emailalert.py
susam: This sounds fair.

I think the function name should be updated. The `_get_connection()` method is not merely getting a connection. It is doing much more than that. It sends the `EHLO` payload behind the scenes. It also performs authentication with `smtp.login()`. This is more of a helper method than something that is merely establishing a connection. We should update the method name to accurately reflect this. Some ideas could be `_start_smtp_session()`, `_prepare_smtp_session()`, `_smtp_session()`, etc.

config.base.yaml
```diff
@@ -29,13 +29,16 @@ alerts:
   emailalert:
     plugin: cloudmarker.alerts.emailalert.EmailAlert
     params:
+      is_ssl_required: True
```
susam: I believe this should be renamed to `use_ssl` considering the parameter name has changed in `EmailAlert` initializer.

cloudmarker/alerts/emailalert.py
```diff
@@ -68,10 +79,38 @@ def done(self):
...
+        apppasswords
+        # 3. In 'select app' choose 'custom', give it an arbitrary name and
+        press generate
+        # 4. It will give you 16 chars token.
+        # 5. Use that token as password for login
+        # 6. Host is smtp.gmail.com and port is 465
+        """
+        # In case no authentication is required then no need to login
+        if self.use_ssl:
+
```
susam: I think we can get rid of this blank line here.

cloudmarker/alerts/emailalert.py
```diff
@@ -68,10 +79,38 @@ def done(self):
...
+        # 1. Log-in into Gmail with your account
+        # 2. Navigate to https://security.google.com/settings/security/
+        apppasswords
+        # 3. In 'select app' choose 'custom', give it an arbitrary name and
+        press generate
+        # 4. It will give you 16 chars token.
+        # 5. Use that token as password for login
+        # 6. Host is smtp.gmail.com and port is 465
+        """
+        # In case no authentication is required then no need to login
```
susam: The `use_ssl` flag should not decide whether authentication should occur or not. Using SSL or not and performing authentication or not are orthogonal concerns. They are independent of each other.

Whether to perform authentication or not should depend on `username` is defined or not. It means, if `username` is defined, then `smtp.login()` should occur in both cases: `use_ssl = True` as well as `use_ssl = False`.

Also, since credentials should never be sent in plaintext, for `use_ssl = False`, we need to invoke `smtp.starttls()` before invoking `smtp.login()`.

cloudmarker/alerts/emailalert.py
```diff
@@ -68,10 +79,53 @@ def done(self):
...
+            smtp_connection.ehlo()
+            smtp_connection.connect(self.host, self.port)
+
+            # an ehlo call is required just after the connect method and just
+            # before login
+            # just to identity yourself. If not called it would throw a
+            # SMTPSenderRefused error
+            # Failed to send email; error: SMTPSenderRefused: (503, b'5.5.1
+            # EHLO/HELO first.
+            smtp_connection.ehlo()
```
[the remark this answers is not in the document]
sunnysharmagts: Done

cloudmarker/alerts/emailalert.py
sunnysharmagts: The `_get_connection()` means a function which returns a smtp connection object. I will rename the function `_prepare_smtp_session()` since it looks more relative to the functionality.

docs/api/cloudmarker.stores.rst
sunnysharmagts: Done

cloudmarker/alerts/emailalert.py
sunnysharmagts: Done

cloudmarker/alerts/emailalert.py
susam: Sounds good.

I am curious if you are also planning to work on #53 (Send email notification after all audits have run) later. If so, this `_prepare_smtp_session()` method has to be moved out to another module outside the plugin packages. The reason is that the change for #53 would be in the framework code (`manager.py` or `workers.py`). The framework cannot depend on individual plugins for its own functionality. So the framework would also have to repeat this work of creating an SMTP connection and starting an SMTP session before it can send email notification related to audits (e.g., an email notification that says that the audits have completed).

So instead of repeating the `_prepare_smtp_session()` code twice (once in `EmailAlert` and once more in the framework code in `manager.py`/`worker.py`), it would make sense to have this `_prepare_smtp_session()` functionality moved to a common place such as `util.py` that both framework and `EmailAlert` can reuse.

Of course, this does not have to be done right now in this pull request. This pull request may remain self contained and not worry about issue #53. What I said above has to be done sooner or later when you or someone else works on #53.

susam:
Amended the commit message to accurately capture the current scope of work (which also includes authentication) and force-pushed the amended commit.

Here is what you would need to do reset your local branch:

```
git checkout master
git pull
git branch -d smtpssl
git checkout smtpssl
```

cloudmarker/alerts/emailalert.py
sunnysharmagts: `SMTP_SSL` connection initiates a TLS encrypted connection. In case we are using a simple `SMTP` connection then in that case we need to encrypt it via `starttls`.

Reference: https://realpython.com/python-send-email/

cloudmarker/alerts/emailalert.py
```diff
@@ -68,10 +77,44 @@ def done(self):
...
+            if self.username:
+                smtp.login(self.username, self.password)
+            return smtp
+
+        smtp = smtplib.SMTP(host=self.host, port=self.port)
+        smtp.set_debuglevel(True)
+        smtp.starttls()
+        if self.username:
+            smtp.login(self.username, self.password)
+        return smtp
```
susam: I see repetition of some statements twice. Unless there is a strong reason, repetition of code should be avoided because it becomes a maintenance overhead in future.

This code can certainly be written without repetition. One example would be:

```python
        if self.use_ssl:
            smtp = smtplib.SMTP_SSL(self.host, self.port)
        else:
            smtp = smtplib.SMTP(self.host, self.port)

        smtp.set_debuglevel(True)
        if not self.use_ssl:
            smtp.starttls()

        if self.username:
            smtp.login(self.username, self.password)

        return smtp
```

Another, perhaps slightly clearer way of writing this with an overhead of a repetition of one statement could be:

```python
        if self.use_ssl:
            smtp = smtplib.SMTP_SSL(self.host, self.port)
            smtp.set_debuglevel(True)
        else:
            smtp = smtplib.SMTP(self.host, self.port)
            smtp.set_debuglevel(True)
            smtp.starttls()

        if self.username:
            smtp.login(self.username, self.password)

        return smtp
```

Note: Please remember to prefix all internal use instance attributes with `_` (underscore) like I mentioned in another comment.

sunnysharmagts: https://github.com/cloudmarker/cloudmarker/pull/75/commits/f72e4709b765dccbf91e95c431acadb93ef30db1#diff-28d4faa43f2d620f5a2c54123370ee47R109

cloudmarker/alerts/emailalert.py
```diff
@@ -68,10 +77,44 @@ def done(self):
...
+        # 1. Log-in into Gmail with your account
+        # 2. Navigate to https://security.google.com/settings/security/
+        apppasswords
+        # 3. In 'select app' choose 'custom', give it an arbitrary name and
+        press generate
+        # 4. It will give you 16 chars token.
+        # 5. Use that token as password for login
+        # 6. Host is smtp.gmail.com and port is 465
+        """
+        # In case no authentication is required then no need to login
```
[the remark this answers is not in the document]
sunnysharmagts: Done

cloudmarker/alerts/emailalert.py
```diff
@@ -14,17 +14,20 @@
...
-
+            subject (str): Subject line for email.
+            to (list): List of recipients (str).
+            sender (str): From field for email.
+            body (str): Email body.
+            use_ssl (bool): Is SSL connection required
+            username (str): username of email account
+            password (str): password of email account
         """
         self.host = host
```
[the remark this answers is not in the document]
sunnysharmagts: https://github.com/cloudmarker/cloudmarker/pull/75/commits/f72e4709b765dccbf91e95c431acadb93ef30db1#diff-28d4faa43f2d620f5a2c54123370ee47R92

cloudmarker/alerts/emailalert.py
```diff
@@ -56,6 +62,9 @@ def done(self):
         this example implementation, we properly terminate the JSON
         array in the email body.
 
+        The connection is made based whether the SMTP or SMTP_SSL is
+        required which is determined by ``username`` param.
+
```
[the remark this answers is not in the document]
sunnysharmagts: https://github.com/cloudmarker/cloudmarker/pull/75/commits/f72e4709b765dccbf91e95c431acadb93ef30db1#diff-28d4faa43f2d620f5a2c54123370ee47R65

cloudmarker/baseconfig.py
```diff
@@ -53,6 +53,26 @@
 alerts:
   filestore:
     plugin: cloudmarker.stores.filestore.FileStore
+  slackalert:
```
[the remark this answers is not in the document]
sunnysharmagts: https://github.com/cloudmarker/cloudmarker/pull/75/commits/f72e4709b765dccbf91e95c431acadb93ef30db1#diff-d90f70f1cb8f6d9b631c6516b682cc04R56

cloudmarker/alerts/emailalert.py
```diff
@@ -56,22 +62,58 @@ def done(self):
...
-        message = MIMEMultipart()
-        message['Date'] = email.utils.formatdate(localtime=True)
-        message['Subject'] = self.subject
-        message['From'] = self.sender
-        message['To'] = email.utils.COMMASPACE.join(self.to)
+        _message = MIMEMultipart()
+        _message['Date'] = email.utils.formatdate(localtime=True)
+        _message['Subject'] = self._subject
+        _message['From'] = self._sender
+        _message['To'] = email.utils.COMMASPACE.join(self._to)
```
susam: These local variables should not begin with an underscore. There is no notion of public vs. internal usage associated with local variables because, well, they are local. Therefore, the convention of beginning an identifier name with underscore does not apply here.

Did you read the style guide I mentioned earlier: [PEP-8](https://www.python.org/dev/peps/pep-0008/)?

It can be confusing to get this right without reading the style guide. I suggest reading the style guide once to understand the coding conventions used in the Python development community. Here are the relevant excerpts:

> ### [Function and Variable Names](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)
> Function names should be lowercase, with words separated by underscores as necessary to improve readability.
>
> Variable names follow the same convention as function names.

> ### [Method Names and Instance Variables](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)
> Use the function naming rules: lowercase with words separated by underscores as necessary to improve readability.
>
> Use one leading underscore only for non-public methods and instance variables.
>
> To avoid name clashes with subclasses, use two leading underscores to invoke Python's name mangling rules.

cloudmarker/alerts/emailalert.py
```diff
@@ -56,22 +62,58 @@ def done(self):
...
+        2. Navigate to https://security.google.com/settings/security/
+        apppasswords
+        3. In 'select app' choose 'custom', give it an arbitrary name and
+        press generate
+        4. It will give you 16 chars token.
+        5. Use that token as password for login
+        6. Host is smtp.gmail.com and port is 465
+
+        """
+        _smtp = None
```
susam: Why is this statement (`_smtp = None`) required? I don't see it serving any purpose. Please remove this statement.

cloudmarker/alerts/emailalert.py
```diff
@@ -56,22 +62,58 @@ def done(self):
...
+        if self._use_ssl:
+            _smtp = smtplib.SMTP_SSL(host=self._host, port=self._port)
+        else:
+            _smtp = smtplib.SMTP(host=self._host, port=self._port)
+            _smtp.starttls()
```
susam: The `smtp.starttls()` call must go after `smtp.set_debuglevel(True)` call otherwise. With the current pull request, we will get no clue about why a session fails if the session fails during `smtp.starttls()`. That partially defeats the purpose of having `smtp.set_debuglevel(True)`.

I have shared two code examples in a previous review to show how to write this code this correctly. Please see [this review comment](https://github.com/cloudmarker/cloudmarker/pull/75#discussion_r276239226).

cloudmarker/alerts/emailalert.py
sunnysharmagts: https://github.com/cloudmarker/cloudmarker/pull/75/commits/d092d76434f27c46450ee99c9fa9ed8ce75f0ba0#diff-28d4faa43f2d620f5a2c54123370ee47R72

cloudmarker/alerts/emailalert.py
sunnysharmagts: https://github.com/cloudmarker/cloudmarker/pull/75/commits/d092d76434f27c46450ee99c9fa9ed8ce75f0ba0#diff-28d4faa43f2d620f5a2c54123370ee47R109

cloudmarker/alerts/emailalert.py
sunnysharmagts: https://github.com/cloudmarker/cloudmarker/pull/75/commits/d092d76434f27c46450ee99c9fa9ed8ce75f0ba0#diff-28d4faa43f2d620f5a2c54123370ee47R109

Issue #78, opened 2019-04-11
Usage Documentation


VigneshSP94:
Is it possible to add usage documentation for this project? Like what this project for and what it does? It will be helpful to get an essence of what it does.

susam:
Hi @VigneshSP94,

Thank you for creating this issue. Yes, we should add usage documentation for this project, so that any potential contributor or user can understand what this project is for, what it does, and how to configure and use it.

The current documentation is lacking in these details. I have created a new branch named `userdoc` and pull request for it (see #79) to address this issue. Visit https://github.com/cloudmarker/cloudmarker/blob/userdoc/README.rst to take a look at what the updated documentation would look like. This is still a work in progress. We are going to add a lot more details about this project in this branch in the next few days.

While we do this, you are also very welcome to contribute to the documentation in particular or the project in general if you wish to do so. If you have any questions about it, please do not hesitate to contact us by creating a [new issue](https://github.com/cloudmarker/cloudmarker/issues/new). Thank you, once again, for bringing our attention to the need for better documentation.

Issue #76, opened 2019-04-11
Rename CLI option -f/--force to -n/--now


susam:
The current CLI options look like this:

```
$ python3 -m cloudmarker -h
usage: cloudmarker [-h] [-c CONFIG [CONFIG ...]] [-f]

optional arguments:
  -h, --help            show this help message and exit
  -c CONFIG [CONFIG ...], --config CONFIG [CONFIG ...]
                        Configuration file paths
  -f, --force           set this flag to force a run
```

I suggest that `-f, --force` be changed to the following:

```
  -n, --now             ignore schedule and run audits now
```

I suggest this change so that the option name makes more sense to new users. The function `cloudmarker.util.parse_cli()` needs to be modified for this.

PR #79, opened 2019-04-11
Add detailed documentation for users (Work in progress)

susam:
Resolves: #78


README.rst
```diff
@@ -13,36 +13,129 @@ Cloudmarker is a cloud security monitoring framework.
...
+   ``rm -rf ~/.venv/cloudmarker`` or enter ``make rmvenv``.
+
 4. Activate the virtual Python environment: ::
 
     . ./venv
 
 5. In the top-level directory of the project, enter this command: ::
 
-    python3 -m cloudmarker
+    python3 -m cloudmarker -n
```
s-nirali: I think you meant `-f`.

susam: @s-nirali I do mean `-n` here. See #76 where I propose that we change this CLI option from `--force/-f` to `--now/-n`. What do you think?

s-nirali: I missed that issue. Makes sense now. `-n` for `now` sounds good.

Issue #14, opened 2019-01-08
Elasticsearch store


s-nirali:
Develop support for Elasticsearch store.

PR #81, opened 2019-04-12
Add EsStore plugin

s-nirali:
The approach to using the buffer in this plugin is slightly different
when compared to other plugins. The documentation on Elasticsearch's
bulk API mentions the following (Ref:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html).

    There is no "correct" number of actions to perform in a single bulk
    call. You should experiment with different settings to find the
    optimum size for your particular workload.

Hence, the buffer size is the maximum number of bytes to be sent in the
HTTP request body of the bulk API.

Resolves: #14

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,8 +2,9 @@ azure-mgmt-compute
 azure-mgmt-network
 azure-mgmt-resource
 azure-mgmt-storage
-pyyaml
-pymongo
+elasticsearch
 google-api-python-client
+pymongo
+pyyaml
 schedule
 slackclient
--- a/cloudmarker/stores/esstore.py
+++ b/cloudmarker/stores/esstore.py
@@ -1,0 +1,118 @@
+"""Elasticsearch store plugin."""
+
+
+import json
+import logging
+
+from elasticsearch import Elasticsearch, ElasticsearchException
+
+_log = logging.getLogger(__name__)
+
+
+class EsStore:
+    """Elasticsearch adapter to index cloud data in Elasticsearch."""
+
+    def __init__(self, host='localhost', port=9200, index='cloudmarker',
+                 buffer_size=5000000):
+        """Create an instance of :class:`EsStore` plugin.
+
+        The plugin uses the default port for Elasticsearch if not
+        specified.
+
+        The ``buffer_size`` for the plugin is the value for the maximum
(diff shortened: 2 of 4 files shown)


susam:
One thing that stands out in this pull request is the decision to define `buffer_size` as the number of bytes to keep in buffer rather the number of records to keep in buffer. This is a deviation from the definition of `buffer_size` in other store plugins.

This deviation seems okay because we normally do not expect users to modify the `buffer_size` parameter of the plugin in configuration. In most cases, the defaults should be good. This deviation appears to be necessary because the Elasticsearch performance appears to suffer based on the number of bytes, not the number of documents. Quoting another reference from Elasticsearch documentation at https://www.elastic.co/guide/en/elasticsearch/guide/current/bulk.html:

> ### How Big Is Too Big?
>
> The entire bulk request needs to be loaded into memory by the node that receives our request, so the bigger the request, the less memory available for other requests. There is an optimal size of bulk request. Above that size, performance no longer improves and may even drop off. The optimal size, however, is not a fixed number. It depends entirely on your hardware, your document size and complexity, and your indexing and search load.
>
> Fortunately, it is easy to find this sweet spot: Try indexing typical documents in batches of increasing size. When performance starts to drop off, your batch size is too big. A good place to start is with batches of 1,000 to 5,000 documents or, if your documents are very large, with even smaller batches.
>
> It is often useful to keep an eye on the physical size of your bulk requests. One thousand 1KB documents is very different from one thousand 1MB documents. A good bulk size to start playing with is around 5-15MB in size.

s-nirali:
@susam , It's a very interesting observation, in your previous comment quoting the "How Big Is Too Big?" passage from the Elasticsearch documentation. I somehow did not stumble upon it while I was looking for optimal sizes because I was only referring to the most recent documentation of ES. Thank you for pointing it out.

requirements.txt
```diff
@@ -2,8 +2,9 @@ azure-mgmt-compute
 azure-mgmt-network
 azure-mgmt-resource
 azure-mgmt-storage
-pyyaml
-pymongo
+elasticsearch
 google-api-python-client
+pymongo
+pyyaml
 schedule
```
susam: I like the sorting of package names here. Makes this file look clean. 👍 🙂

Issue #89, opened 2019-04-14
Update command to run audits right now


susam:
Point 5 in the instructions provided in the README contains this command:

```shell
python3 -m cloudmarker
```

This does not run audits right now but instead waits until a scheduled time to start audits. This could be confusing to someone new to the project. Replace this command with the following:

```shell
python3 -m cloudmarker -n
```

This command runs the audits right now.

PR #90, opened 2019-04-14
Search for user configuration in popular locations

susam:
Since Cloudmarker is supposed to serve as a CLI tool that one can run
with `cloudmarker -n` as well as a service that one may run as
`cloudmarker`, it is convenient if Cloudmarker searches certain popular
paths where configuration files usually reside. This would prevent the
users or service unit definition from having to specify configuration
files as arguments to the `-c` or `--config` option.

With this change, it now looks for a configuration file at the following
locations in the order specified:

  - /etc/cloudmarker.yaml
  - ~/.cloudmarker.yaml
  - ~/cloudmarker.yaml
  - cloudmarker.yaml (i.e., in the current working directory)

The base configuration file (which was named `config.base.yaml`) earlier
has been removed from the file system. There is no counterpart for it
now on the file system. Instead the base configuration is now built into
the code in a new module named `baseconfig`. This has been done for the
following reasons:

  - Provide reasonable defaults as built-in. This is consistent with how
    other tools work, e.g., Vim, Bash, etc. ship with defaults and we
    write `~/.vimrc`, `~/.bashrc`, etc. only when we want to override
    these defaults.

  - The users were never supposed to edit `config.base.yaml` anyway. By
    removing it from the file system, the users do not have to worry
    about notions like base configuration and user configuration. They
    do not have to be careful about not editing the base configuration.
    The users can still refer to the base configuration if they wish to
    by entering `cloudmarker --print-base-config`.

Since multiple configuration paths are involved, the logging has been
improved to show what configuration paths are being looked for and which
configuration paths are being found.

Resolves: #89

### DIFF
--- a/cloudmarker/__init__.py
+++ b/cloudmarker/__init__.py
@@ -1,5 +1,5 @@
 """Cloudmarker - Cloud security monitoring framework."""
 
 
-__version__ = '0.1.0.dev1'
+__version__ = '0.1.0.dev2'
 __author__ = 'Cloudmarker Authors and Contributors'
--- a/cloudmarker/baseconfig.py
+++ b/cloudmarker/baseconfig.py
@@ -1,3 +1,37 @@
+"""Base configuration.
+
+Attributes:
+    config_yaml (str): Base configuration as YAML code.
+    config_dict (dict): Base configuration as Python dictionary.
+
+Examples:
+    Here are a few examples that show how the values of these attributes
+    look::
+
+        >>> from cloudmarker import baseconfig
+        >>> print(baseconfig.config_yaml) # doctest: +ELLIPSIS
+        # Base configuration
+        clouds:
+          mockcloud:
+            plugin: cloudmarker.clouds.mockcloud.MockCloud
+        ...
+        >>> baseconfig.config_dict['clouds']
+        {'mockcloud': {'plugin': 'cloudmarker.clouds.mockcloud.MockCloud'}}
+        >>> baseconfig.config_dict['audits'] # doctest: +ELLIPSIS
+        {'mockaudit': {...}}
+        >>> baseconfig.config_dict['audits']['mockaudit']['clouds']
+        ['mockcloud']
+
+    .. Note that it is necessary to put the above example in a
+       reStructuredText literal code block created with the "::" marker
+       so that the doctest directive "# doctest: +ELLIPSIS" does not
+       appear in the rendered documentation.
+
+"""
+
+import yaml
+
+config_yaml = """# Base configuration
 clouds:
   mockcloud:
     plugin: cloudmarker.clouds.mockcloud.MockCloud
@@ -11,31 +45,14 @@ stores:
     plugin: cloudmarker.stores.mongodbstore.MongoDBStore
 
 events:
+  firewallruleevent:
+    plugin: cloudmarker.events.firewallevent.FirewallRuleEvent
   mockevent:
     plugin: cloudmarker.events.mockevent.MockEvent
-  firewallevent:
-    plugin: cloudmarker.evengts.firewallevent.FirewallEvent
 
 alerts:
   filestore:
     plugin: cloudmarker.stores.filestore.FileStore
-  slackalert:
-    plugin: cloudmarker.alerts.slackalert.SlackAlert
-    params:
-      bot_user_token: foo-bot-token
-      to:
-        - foo@example.com
-      text: foo
-  emailalert:
-    plugin: cloudmarker.alerts.emailalert.EmailAlert
-    params:
-      host: smtp.example.com
-      port: 25
-      sender: CloudMarker Alert <no-reply@example.com>
-      to:
-        - recipient@example.com
-      subject: Anomaly notification
-      body: default text
 
 audits:
   mockaudit:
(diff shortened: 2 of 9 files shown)


cloudmarker/__init__.py
```diff
@@ -1,5 +1,5 @@
 """Cloudmarker - Cloud security monitoring framework."""
 
 
-__version__ = '0.1.0.dev1'
+__version__ = '0.1.0.dev2'
```
susam: The version number has been incremented because I would like to make another dev release of Cloudmarker on PyPI after this PR is merged, so that I can install the new version using `pip3 install cloudmarker` just like a user would.

This would help me to test out the instructions I would be writing for users in the new README currently being worked upon at [userdoc:README.rst](https://github.com/cloudmarker/cloudmarker/blob/userdoc/README.rst).

cloudmarker/test/test_main.py
```diff
@@ -8,7 +8,8 @@
 class MainTest(unittest.TestCase):
     """Tests for package execution."""
 
-    @mock.patch('sys.argv', ['cloudmarker', '-c', 'config.base.yaml', '-n'])
+    @mock.patch('sys.argv', ['cloudmarker', '-n', '-c',
+                             'cloudmarker.base.yaml'])
```
s-nirali: This works without the 4th argument `cloudmarker.base.yaml`.

cloudmarker/test/test_util.py
```diff
@@ -71,6 +80,28 @@ def test_parse_cli_args_config(self):
         args = util.parse_cli(['--config', 'baz.yaml', 'qux.yaml'])
         self.assertEqual(args.config, ['baz.yaml', 'qux.yaml'])
```
s-nirali: I think there should also be a test for default config with just `-c` option like

```
    def test_parse_cli_args_no_config(self):
        # Short option.
        args = util.parse_cli(['-c'])
        self.assertEqual(args.config, [])
```

cloudmarker/util.py
```diff
@@ -118,18 +126,85 @@ def parse_cli(args=None):
...
+    Missing config files are ignored.
+
+    If two or more config files provide conflicting config values, the
+    config file specified later overrides the built-in base config and
+    the config files specified earlier.
+
+    If the -c/--config option is specified without any file arguments
+    following it, then only the built-in base config is used.
+
+    If the -c/--config option is omitted, then the following config
```
mitprasoon: If -c/--config option is specified with file arguments then the code tries to read those files and load the configuration from them. In case the file is non-existent the file is ignored. Shouldn't it gracefully exit in case a user specified file is missing?  

Whereas, if -c/--config option is omitted it should look for config files in popular location, load and merge them with the built-in base config.

cloudmarker/baseconfig.py
```diff
@@ -86,3 +104,7 @@
       - file_handler
 
 schedule: "00:00"
+"""
+
+
+config_dict = yaml.load(config_yaml)
```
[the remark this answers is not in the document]
susam: I like the current approach because of three reasons:

- By using `config_yaml` string as the original source of base config, the `--print-base-config` option can print the YAML keys in the order we intend. `yaml.dump(config_dict)` may print the keys in any arbitrary order which may not aid comprehensibility.
- We can put comments in the `config_yaml` string and they would appear intact with the `--print-base-config` option.
- Any custom formatting (such as blank lines etc.) we use to aid readability would appear intact with the `--print-base-config` option. This is same as the point you mentioned.

cloudmarker/util.py
```diff
@@ -118,18 +126,85 @@ def parse_cli(args=None):
...
+                        help='print base configuration')
+
     parser.add_argument('-v', '--version', action='version',
                         version='%(prog)s ' + cloudmarker.__version__)
+
     args = parser.parse_args(args)
     return args
 
 
+def wrap_paragraphs(text, width=70):
```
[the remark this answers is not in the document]
susam: I don't have a strong opinion about this. But right now, I have a preference for it being in the same `util` module. I like that all utilities are in a single `util` module and we do not have to think which utility function belongs to which `util` module.

Also, another way to look at this is that `wrap_paragraphs` is a generic utility function and `util.py` does not know who wants to use it and for what purpose. Right now we are using it for the `--help` message but later we could also reuse it to format some other kind of output or alert message.

cloudmarker/util.py
susam: This sounds like a good idea. This will add a little bit of complexity to the `parse_cli()` and `load_config()` functions, so I will create a separate pull request for it, so that we can discuss it separately and see if this feature is important enough to introduce this little bit of complexity.

cloudmarker/test/test_main.py
susam: Good catch. [Fixed here](https://github.com/cloudmarker/cloudmarker/pull/90/files#diff-3cfc46b1f4ad6f67f74d79216e9c6eb5R11).

cloudmarker/test/test_util.py
susam: Yes, I like this suggestion. I have added a new test method here: [test_parse_cli_args_no_config](https://github.com/cloudmarker/cloudmarker/pull/90/files#diff-45528363d1b141abd9b51e1eb4f6e297R75).

PR #96, opened 2019-04-18
Add tests for SplunkHECStore plugin

r0hi7:


cloudmarker/test/test_splunkhecstore.py
```diff
@@ -0,0 +1,120 @@
...
+
+        splunk_store._flush = mock.MagicMock()
+
+        splunk_store.write(mock_record)
+        splunk_store.done()
+
+        self.assertEqual(splunk_store._flush.mock_calls,
+                         [mock.call(), mock.call()])
+
+    @mock.patch('cloudmarker.stores.splunkhecstore.requests.session')
```
susam: It is sufficient to only `@mock.patch('requests.session')`. It is unnecessary to get access to `requests.session` by first referring to the `cloudmarker.stores.splunkhecstore` module in this case.

If we were to mock `requests` package, then we would have needed the longer form of patching, i.e., `@mock.patch('cloudmarker.stores.splunkhecstore.requests')`. Why? Because `@mock.patch('requests')` would not affect the `requests` package that has already been imported into `splunkhecstore`. In this test module, we first `import splunkhecstore` which in turn imports `requests` inside `splunkhecstore`, so `@mock.patch('requests')` in the test module does not change the `requests` package that is already imported in `splunkhecstore`. The situation looks like this:

```
Before @mock.patch('requests'):
    in test_splunkhecstore:
        requests --- refers to ---> <actual requests package object in memory>
        requests.session --- refers to ---> <actual session function in memory>
    in splunkhecstore:
        requests --- refers to ---> <the same requests package object in memory>
        requests.session --- refers to ---> <the same session function in memory>

After @mock.patch('requests'):
    in test_splunkhecstore:
        requests --- now refers to ---> (a MagicMock object in memory)
        requests.session --- now refers to ---> (a dynamically created MagicMock object in memory)
    in splunkhecstore:
        requests --- still refers to ---> <actual requests package object in memory>
        requests.session --- still refers to ---> <actual session function in memory>
```

Note that in the beginning, both `test_splunkhecstore` and `splunkhecstore` refer to the same `requests` package object imported in memory. That's because a package/module is imported only once in the lifetime of a project regardless of how many times we have imported them in different parts of the project with the `import` statement.

Then note that after mocking, the `splunkhecstore`'s `requests` reference is still referring to the original `requests` even after mocking.

But in this test module since, we are mocking `requests.session` here, not `requests`, the situation turns out to be different:

```
Before @mock.patch('requests.session'):
    in test_splunkhecstore:
        requests --- refers to ---> <actual requests package object in memory>
        requests.session --- refers to ---> <actual session function in memory>
    in splunkhecstore:
        requests --- refers to ---> <the same requests package object in memory>
        requests.session --- refers to ---> <the same session function in memory>

After @mock.patch('requests.session'):
    test_splunkhecstore:
        requests --- still refers to ---> <actual requests package object in memory>
        requests.session --- now refers to ---> (a MagicMock object in memory)
    in splunkhecstore:
        requests --- still refers to ---> <actual requests package object in memory>
        requests.session --- now refers to ---> (a MagicMock object in memory)
```

With `mock.patch('requests.session')`, in the test module, we are referring to the same `requests` package object that `splunkhecstore` is also referring to, so when we replace its `session` function, i.e., `requests.session` with a mock, `splunkhecstore` also gets the same mock when it tries to use `requests.session`.

Summary: We can use just `@mock.patch('requests.session')` in this test module.

cloudmarker/test/test_splunkhecstore.py
```diff
@@ -0,0 +1,120 @@
...
+        # In case of post request to splunk fails, buffer should not be empty.
+        self.assertNotEqual(splunk_store._buffer, [])
+
+    @mock.patch('cloudmarker.stores.splunkhecstore.requests.session')
+    def test_flush_splunk_unable_to_index(self, mock_session):
+        mock_record = {'record_type': 'firewall_rule'}
+
+        mock_response = mock.MagicMock()
+        mock_response.status_code = 200
+        mock_response.json.return_value = {'code': 1}
```
susam: In this case, I believe the mock return value has to be something like `{'code': 1, 'text': 'foo'}`. Otherwise, we are getting a `KeyError: 'text'` in line 87 of `splunkhecstore.py` due to which line 88 is never reached. As a result, it is preventing 100% code coverage for `splunkhecstore.py`.

pylama.ini
```diff
@@ -40,14 +40,15 @@ ignore = W0703
...
-ignore = D102,C0111,R0903,R0904,R0201,W0613
+ignore = D102,C0111,R0903,R0904,R0201,W0613,W0212
 
 # D102 Missing docstring in public method [pydocstyle]
 # C0111 Missing method docstring [pylint]
 # R0903 Too few public methods (0/2) [pylint]
 # R0904 Too many public methods (23/20) [pylint]
 # R0201 Method could be a function [pylint]
 # W0613 Unused argument 'mock_service_account' in test_gcpcloud.py
+# W0212 Access to a protected member _flush of a client class
```
susam: This warning should not be ignored. This is a good warning. We should not access internal member of a class from outside. Once the tests are updated like I explained in my review comments, there would be no need to ignore this warning.

r0hi7:
Thanks for the feedback. Give me some time i will update the PR :)

cloudmarker/test/test_splunkhecstore.py
r0hi7: 👍

PR #97, opened 2019-04-18
Add tests for EsStore plugin

r0hi7:


cloudmarker/test/test_esstore.py
```diff
@@ -0,0 +1,73 @@
...
+                }
+            ]
+        }
+
+        mock_bulk.return_value = mock_bulk_response
+
+        es_store.write(mock_record)
+        es_store.done()
+
+        mock_bulk.assert_called_once()
```
s-nirali: I'd mock the logger here as well.

```
@mock.patch('cloudmarker.stores.esstore._log')
    def test_bulk_response_error(self, mock_logger, mock_bulk):
```

```
mock_logger.debug.assert_called_with('Failed to insert record; ID: %s', 'foo')
mock_logger.error.assert_called_with('Failed to write %d records', 1)
```

cloudmarker/test/test_esstore.py
```diff
@@ -0,0 +1,73 @@
...
+
+        es_store = esstore.EsStore()
+
+        mock_bulk.side_effect = ElasticsearchException()
+
+        es_store.write(mock_record)
+        es_store.done()
+
+        # In case of any expception the buffer should not reset.
+        self.assertNotEqual(es_store._buffer, '')
```
s-nirali: Not sure if just checking if buffer is empty would be enough to ensure an exception was raised. I'd rather check if the error was logged. Maybe by mocking the logger itself like below.

```
@mock.patch('cloudmarker.stores.esstore._log')
    def test_exception_flow(self, mock_logger, mock_bulk):
```

and then asserting the logs

```
mock_logger.error.assert_called_with('Bulk Index Error: %s: %s', 'ElasticsearchException', mock_bulk.side_effect)
```

cloudmarker/test/test_esstore.py
```diff
@@ -0,0 +1,73 @@
...
+            'errors': '',
+            'items': [
+                {'index': {'status': 200}}
+            ]
+        }
+
+        mock_bulk.return_value = mock_bulk_response
+
+        es_store.write(mock_record)
+        es_store.done()
```
s-nirali: After the `write`, I'd check if the buffer actually got filled though that's up to your discretion if you want it here or not. This test works well and asserting the buffer after the write is just for more clarity.

cloudmarker/test/test_esstore.py
```diff
@@ -0,0 +1,73 @@
...
+        self.assertNotEqual(es_store._buffer, '')
+
+    @mock.patch('cloudmarker.stores.esstore.Elasticsearch.bulk')
+    def test_bulk_response_error(self, mock_bulk):
+        mock_record = {'record_type': 'firewall_rule'}
+
+        es_store = esstore.EsStore()
+
+        mock_bulk_response = {
+            'errors': 'foo',
```
s-nirali: Also forgot to mention, the `errors` key here is supposed to have a value of `true` or `false` though it doesn't affect or break the test in any way and having the value to be `foo` is also considered as `true`.

cloudmarker/test/test_esstore.py
```diff
@@ -0,0 +1,73 @@
...
+from elasticsearch import ElasticsearchException
+
+from cloudmarker.stores import esstore
+
+
+class EsStoreTest(unittest.TestCase):
+    """Tests for EsStore plugin."""
+
+    @mock.patch('cloudmarker.stores.esstore.Elasticsearch.bulk')
+    def test_happy_flow(self, mock_bulk):
```
susam: I think we can make the `@mock.patch()` simpler.

- Update `esstore.py` to `import elasticsearch`.
- Then in these test methods, `@mock.patch('esstore.Elasticsearch.bulk')`.

I guess this is one more reason to follow the convention of importing modules (not classes and functions) in the application code. It makes it easier to patch them with a mock in the test methods.

cloudmarker/test/test_esstore.py
susam: I doubt there is a way to check if the buffer got filled after `write()` without relying on testing the internal state. Asserting on an internal attribute breaks encapsulation and tightly couples the test methods with the internal implementation of a class, so it should be avoided. I discussed this in some detail [here](https://github.com/cloudmarker/cloudmarker/pull/96#discussion_r277122687).

What we should test instead is whether the public methods of Elasticsearch were invoked like we expect with the help of mocks. This has been done below in line 32, so I think this test method is good.

One might argue how Elasticsearch is used in a class is also an internal implementation detail. The answer lies in the publicly visible behaviour we want to test. We do not want to test how `EsStore` internally manages its buffer. We only to test that if `EsStore`'s `write()` and `done()` methods are called, then we hand over the records to Elasticsearch. To test this it becomes necessary to at least ensure that `Elasticsearch`'s `bulk()` method was called. That's the maximum extent of how far we should go regarding testing internal implementation of a class.

cloudmarker/test/test_esstore.py
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        mock_bulk.return_value = mock_bulk_response
+
+        es_store.write(mock_record)
+        es_store.done()
+
+        mock_bulk.assert_called_once()
+        # In case of successful push the buffer should reset.
+        self.assertEqual(es_store._buffer, '')
```
susam: Please remove this assert. The tests should not care how `EsStore` manages its buffer. It is an internal implementation detail.

cloudmarker/test/test_esstore.py
susam: I agree with @s-nirali. I would suggest first getting rid of asserting on `_buffer` because it is an internal attribute of the class. That's sufficient for this pull request. The fact that the class does not crash due to an `ElasticsearchException` is a valuable test in itself.

Then, if you want to go ahead and assert the logging in this pull request or in a future one, then I would suggest something like this so that the test method is not tightly coupled with the exact error message being logged:

```
self.assertEqual(len(mock_logger.error.mock_calls), 1)
```

cloudmarker/test/test_esstore.py
susam: I suggest not testing info and debug logs. The logs are meant for diagnostics which is a bonus addition to the documented behaviour of the project. I prefer not testing these bonus additions. Having said that, I think there is value in testing warning and error logs for exceptional conditions (such as an exception). Logging warning or error for exceptional conditions is a necessary feature and we want to be sure that these logs accidentally do not disappear in future commits, so having tests for these sounds like a good idea.

pylama.ini
```diff
@@ -40,14 +40,15 @@ ignore = W0703
...
-ignore = D102,C0111,R0903,R0904,R0201,W0613
+ignore = D102,C0111,R0903,R0904,R0201,W0613,W0212
 
 # D102 Missing docstring in public method [pydocstyle]
 # C0111 Missing method docstring [pylint]
 # R0903 Too few public methods (0/2) [pylint]
 # R0904 Too many public methods (23/20) [pylint]
 # R0201 Method could be a function [pylint]
 # W0613 Unused argument 'mock_service_account' in test_gcpcloud.py
+# W0212 Access to a protected member _buffer of a client class
```
susam: Once the tests are updated to not rely on internal members, we can get rid of this change.

cloudmarker/test/test_esstore.py
s-nirali: @susam , is it okay to mock logger since `_log` is also an internal module attribute?

susam: @s-nirali You are right. We should not patch `_log` because it is an internal attribute. If we want to test it the right way, then we would have to expose a public function, say, `setLogger()` in the module that would set its internal `_log` variable to a new logger and then use this function to set the logger to a mock. I think this is an overkill just for the purpose of testing. Another option is to not care that `_log` is internal and patch it anyway. Yet another option is to not bother about testing logger calls.

cloudmarker/test/test_esstore.py
r0hi7: Can we find some other ways to test to identify if the exceptions are raised or not? Checking the values for the loggers is not very much convincing to me. I also initially thought the same.

cloudmarker/test/test_esstore.py
r0hi7: Then what is the best assertion that falls under this? I am still confused about what should I choose the assert statement to be?
I am with @susam  on this about not using the internal state of the class attributes.

r0hi7:
@susam , @s-nirali i have updated the PR. I have made minor changes. Please review.

cloudmarker/test/test_esstore.py
s-nirali: @r0hi7 feel free to update import paths in `esstore.py` so that the patches in this file could be made simpler.

r0hi7: Can I do it in a separate commit? As this would also require to change the esstore.py file.

susam: I would have preferred the `esstore.py` change in this commit itself. The `esstore.py` change should be straightforward. However, if you prefer to do it in a separate commit, that's fine too.

cloudmarker/test/test_esstore.py
susam: If we have `mock_bulk.side_effect = ElasticsearchException()` in the test code, then the exception is definitely being raised. There is no need to test if the exception is raised. That would amount to only testing whether `unittest.mock` is working as expected. We don't want to test `unittest.mock`. We want to test `EsStore`.

There is no need to explicitly test the exception here because the raising of exception and handling it is internal behavior in the current `EsStore`. The external behavior is just that `EsStore` does not crash which the existing tests are testing. There is of course logging as another external behaviour but I guess we are deciding here not to test that.

Finally, how can we be sure that our test itself is correct and it did trigger an exception which was handled by the `except` block? For that, we have code coverage. For example, see this [code coverage report for esstore.py](https://coveralls.io/builds/22962027/source?filename=cloudmarker/stores/esstore.py) triggered due to this pull request. It shows that lines 70-74, i.e., the `except` block were exercised by our test. So the code coverage report proves that our tests were able to trigger an exception that `EsStore` could handle.

PR #98, opened 2019-04-22
Make EmailAlert docs and code consistent

susam:
Make EmailAlert docs and code consistent

The EmailAlert docstrings have been updated to make it look good from a
user's perspective. In some places, the docstrings have been fixed to
describe the functionality in the code correctly. In some other cases,
the code has been updated to be consistent with the updated docstrings.

The following changes have been made to the plugin parameters:

  - Rename `sender` to `from_addr`.
  - Rename `to` to `to_addrs`.
  - Remove `body`.
  - Add defaults for `subject`, `host`, and `port`.

The names `from_addr` and `to_addrs` have been picked from the Python
`smtplib` documentation. They have done the work of choosing good names,
so we just reuse it. The `to_addrs` name has an additional advantage
that it hints at the value being a list of addresses.

The `body` parameter has been removed because this is not a generic
emailer class. This is an alert plugin which will decide its own email
content based on the events it receives in its `write()` method.
There is never a situation where the email content can be specified via
configuration.

The defaults for `host` and `port` have also been picked from the Python
`smtplib` documentation. They implement sensible default behaviour for
these defaults, so we reuse that behaviour by choosing the same
defaults.

### DIFF
--- a/cloudmarker/alerts/emailalert.py
+++ b/cloudmarker/alerts/emailalert.py
@@ -12,98 +12,86 @@ _log = logging.getLogger(__name__)
 
 
 class EmailAlert:
-    """A plugin to send email notification for anomalies found."""
+    """A plugin to send email alerts."""
 
-    def __init__(self, host, port, subject, to, sender, body, use_ssl=True,
+    def __init__(self, from_addr, to_addrs,
+                 subject='Cloudmarker Alert',
+                 host='', port=0, use_ssl=True,
                  username=None, password=None):
         """Create an instance of :class:`EmailAlert` plugin.
 
+        When ``use_ssl`` is ``True`` and ``host`` is not specified or
+        specified as ``''``, the local host is used. When ``use_ssl`` is
+        ``True`` and ``port`` is not specified or specified as ``0``,
+        the standard SMTP-over-SSL, i.e., port 465, is used. See
+        :class:`smtplib.SMTP_SSL` documentation for more details on
+        this.
+
+        When ``use_ssl`` is ``False`` and if ``host`` or ``port`` are not
+        specified, i.e., if host or port are ``''`` and/or ``0``
+        respectively, the OS default behavior is used. See
+        :class:`smtplib.SMTP` documentation for more details on this.
+
+        If ``username`` is not specified or specified as ``None``, no
+        SMTP authentication is done. If ``username`` is specified and it
+        is not ``None``, then SMTP authentication is done.
+
         Arguments:
-            host (str): Server hostname for SMTP.
-            port (int): Server port for SMTP.
-            subject (str): Subject line for email.
-            to (list): List of recipients (str).
-            sender (str): From field for email.
-            body (str): Email body.
-            use_ssl (bool): Is SSL connection required
-            username (str): username of email account
-            password (str): password of email account
+            from_addr (str): Sender's email address.
+            to_addrs (list): A list of :obj:`str` objects where each
+                :obj:`str` object is a recipient's email address.
+            host (str): SMTP host.
+            port (int): SMTP port.
+            use_ssl (bool): Use SSL if ``True``, not otherwise.
+            username (str): SMTP username.
+            password (str): SMTP password.
         """
         self._host = host
         self._port = port
         self._subject = subject
-        self._to = to
-        self._sender = sender
-        self._body = body
+        self._from_addr = from_addr
+        self._to_addrs = to_addrs
         self._stringbuffer = []
         self._use_ssl = use_ssl
         self._username = username
         self._password = password
 
     def write(self, record):
-        """Write JSON records to the file system.
-
-        This method is called once for every ``record`` read from a
-        cloud. In this example implementation of a alert, we simply
-        send the ``record`` in JSON format via email to the recipient.
-        The records keeps appending to a stringbuffer which then is
-        written in the email body
+        """Save event record in a buffer.
 
         Arguments:
-            record (dict): Data to send via email.
+            record (dict): An event record.
         """
         for _, value in record.items():
             self._stringbuffer.append(repr(value))
 
     def done(self):
-        """Perform final cleanup tasks.
-
-        This method is called after all records have been written. In
-        this example implementation, we properly terminate the JSON
-        array in the email body.
-
-        The connection is made based whether the SMTP or SMTP_SSL is
-        required which is determined by ``use_ssl`` param. The SMTP
-        login process is conducted if ``username`` param is provided.
-
-        """
+        """Send the buffered events as an email alert."""
         message = MIMEMultipart()
         message['Date'] = email.utils.formatdate(localtime=True)
         message['Subject'] = self._subject
-        message['From'] = self._sender
-        message['To'] = email.utils.COMMASPACE.join(self._to)
+        message['From'] = self._from_addr
+        message['To'] = email.utils.COMMASPACE.join(self._to_addrs)
 
-        # Incase if the buffer is empty then the default message will be sent
-        if self._stringbuffer:
-            self._body = ''.join(self._stringbuffer)
-        message.attach(MIMEText(self._body))
+        body = ''.join(self._stringbuffer)
+        message.attach(MIMEText(body))
 
         smtp = self._prepare_smtp_session()
         try:
-            smtp.sendmail(self._sender, self._to, message.as_string())
+            smtp.sendmail(self._from_addr,
+                          self._to_addrs,
+                          message.as_string())
         except smtplib.SMTPException as e:
             _log.error('Failed to send email: %s', e)
         finally:
             smtp.quit()
 
     def _prepare_smtp_session(self):
-        """Return SMTP connection object.
+        """Start an SMTP session.
 
-        Create a SMTP connection based on whether the SSL param.
-        If SSL connection is required (for eg: in case of gmail) then
-        method will return a SMTP_SSL connection. In other cases if the
-        authentication is not required or SMTP_SSL connection is not
-        required then a plain SMTP connection object is returned.
-        For gmail follow the steps given below even to work for 2 factor
-        authentication
-        1. Log-in into Gmail with your account
-        2. Navigate to https://security.google.com/settings/security/
-        apppasswords
-        3. In 'select app' choose 'custom', give it an arbitrary name and
-        press generate
-        4. It will give you 16 chars token.
-        5. Use that token as password for login
-        6. Host is smtp.gmail.com and port is 465
+        Returns:
+            smtplib.SMTP or smtplib.SMTP_SSL: An SMTP connection
+                with its session ready to send messages.
 
         """
         if self._use_ssl:


cloudmarker/alerts/emailalert.py
```diff
@@ -12,98 +12,86 @@
...
         """Create an instance of :class:`EmailAlert` plugin.
 
+        When ``use_ssl` is ``True`` and ``host`` is not specified or
+        specified as ``''``, the local host is used. When ``use_ssl`` is
+        ``True`` and ``port`` is not specified or specified as ``0``,
+        the standard SMTP-over-SSL, i.e., port 465, is used. See
+        :class:`smtplib.SMTP_SSL` documentation for more details on
+        this.
+
+        When ``use_ssl`` is ``False` and if ``host`` or ``port`` are not
```
s-nirali: Missing backtick

cloudmarker/alerts/emailalert.py
```diff
@@ -12,98 +12,86 @@
...
+    """A plugin to send email alerts."""
 
-    def __init__(self, host, port, subject, to, sender, body, use_ssl=True,
+    def __init__(self, from_addr, to_addrs,
+                 subject='Cloudmarker Alert',
+                 host='', port=0, use_ssl=True,
                  username=None, password=None):
         """Create an instance of :class:`EmailAlert` plugin.
 
+        When ``use_ssl` is ``True`` and ``host`` is not specified or
```
s-nirali: Missing backtick

cloudmarker/alerts/emailalert.py
```diff
@@ -12,98 +12,86 @@
...
-        message.attach(MIMEText(self._body))
+        body = ''.join(self._stringbuffer)
+        message.attach(MIMEText(body))
 
         smtp = self._prepare_smtp_session()
         try:
-            smtp.sendmail(self._sender, self._to, message.as_string())
+            smtp.sendmail(self._from_addr,
+                          self._to_addrs,
+                          message.as_string())
```
s-nirali: Most of the other plugins have a success message logged which I personally find very convenient to verify if the method actually succeeded. The other alert, however, does not have it as well. You may choose to add it if you feel it is required.

cloudmarker/alerts/emailalert.py
```diff
@@ -12,98 +12,86 @@
...
-        5. Use that token as password for login
-        6. Host is smtp.gmail.com and port is 465
+        """Start an SMTP session.
+
+        Returns:
+            smtplib.SMTP or smtplib.SMTP_SSL: An SMTP connection
+                with its session ready to send messages.
 
         """
         if self._use_ssl:
```
s-nirali: In lines [99](https://github.com/cloudmarker/cloudmarker/pull/98/commits/ba56330d1b85174f7258121ce04b481267580cec#diff-28d4faa43f2d620f5a2c54123370ee47R99) and [102](https://github.com/cloudmarker/cloudmarker/pull/98/commits/ba56330d1b85174f7258121ce04b481267580cec#diff-28d4faa43f2d620f5a2c54123370ee47R102), enabling the debug level by default leads to a lot of email body data in the logs. Should this flag be moved to the config as an internal debug parameter?

cloudmarker/alerts/emailalert.py
```diff
@@ -12,98 +12,86 @@
...
-        message['From'] = self._sender
-        message['To'] = email.utils.COMMASPACE.join(self._to)
+        message['From'] = self._from_addr
+        message['To'] = email.utils.COMMASPACE.join(self._to_addrs)
 
-        # Incase if the buffer is empty then the default message will be sent
-        if self._stringbuffer:
-            self._body = ''.join(self._stringbuffer)
-        message.attach(MIMEText(self._body))
+        body = ''.join(self._stringbuffer)
```
s-nirali: Is it okay to send raw json data in the email body?

cloudmarker/alerts/emailalert.py
s-nirali: At line [98](https://github.com/cloudmarker/cloudmarker/pull/98/commits/ba56330d1b85174f7258121ce04b481267580cec#diff-28d4faa43f2d620f5a2c54123370ee47R98), with the following config, I ran into an `ssl.SSLError`.

Config:
```json
alerts:
  emailalert:
    params:
      use_ssl: True
      host: smtp.gmail.com
      port: 587
      from_addr: CloudMarker Alert <no-reply@gmail.com>
      to_addrs:
        - <recipient@example.com>
      subject: Anomaly notification
      username: <alice>
      password: <alice@123>
```

Exception stack trace:
```
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/git/cloudmarker/cloudmarker/workers.py", line 68, in store_worker
    store_plugin.done()
  File "/Users/git/cloudmarker/cloudmarker/alerts/emailalert.py", line 79, in done
    smtp = self._prepare_smtp_session()
  File "/Users/git/cloudmarker/cloudmarker/alerts/emailalert.py", line 98, in _prepare_smtp_session
    smtp = smtplib.SMTP_SSL(host=self._host, port=self._port)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/smtplib.py", line 1031, in __init__
    source_address)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/smtplib.py", line 251, in __init__
    (code, msg) = self.connect(host, port)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/smtplib.py", line 336, in connect
    self.sock = self._get_socket(host, port, self.timeout)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/smtplib.py", line 1039, in _get_socket
    server_hostname=self._host)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ssl.py", line 412, in wrap_socket
    session=session
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ssl.py", line 853, in _create
    self.do_handshake()
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ssl.py", line 1117, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLError: [SSL: UNKNOWN_PROTOCOL] unknown protocol (_ssl.c:1056)
```

If the `use_ssl` key is set to `False`, the mail is successfully sent. This could probably be handled in another PR. There is a solution mentioned [here](https://stackoverflow.com/a/49443966).

susam:
>     2. Should the buffer be checked for an upper limit like 25 Mb by default for size?

Hi Nirali,

We can do something like this. I would prefer to do this in a separate pull request. By the way, did you come up with the 25 MB limit as an arbitrary limit or did you use some reference to come up with this limit?

cloudmarker/alerts/emailalert.py
susam: I agree. I was trying to keep this change minimal, i.e., just focus on the documentation and necessary code updates. This plugin requires quite a bit of rework to move the SMTP code out to `util` module and make it consistent with the other plugins. I will do some of it as part of #53. I will address this suggestion of yours while resolving #53.

cloudmarker/alerts/emailalert.py
susam: > In lines [99](https://github.com/cloudmarker/cloudmarker/pull/98/commits/ba56330d1b85174f7258121ce04b481267580cec#diff-28d4faa43f2d620f5a2c54123370ee47R99) and [102](https://github.com/cloudmarker/cloudmarker/pull/98/commits/ba56330d1b85174f7258121ce04b481267580cec#diff-28d4faa43f2d620f5a2c54123370ee47R102), enabling the debug level by default leads to a lot of email body data in the logs. Should this flag be moved to the config as an internal debug parameter?

I agree. This should be config driven. I will do this as part of #53.

s-nirali:
> > ```
> > 2. Should the buffer be checked for an upper limit like 25 Mb by default for size?
> > ```
> 
> Hi Nirali,
> 
> We can do something like this. I would prefer to do this in a separate pull request. By the way, did you come up with the 25 MB limit as an arbitrary limit or did you use some reference to come up with this limit?

I used this, https://www.outlook-apps.com/maximum-email-size/.

cloudmarker/alerts/emailalert.py
susam: > At line [98](https://github.com/cloudmarker/cloudmarker/pull/98/commits/ba56330d1b85174f7258121ce04b481267580cec#diff-28d4faa43f2d620f5a2c54123370ee47R98), with the following config, I ran into an `ssl.SSLError`.
> 
> Config:
> 
> ```json
> alerts:
>   emailalert:
>     params:
>       use_ssl: True
>       host: smtp.gmail.com
>       port: 587

For `use_ssl` set to `True`, set `port` to either `465` or `0` (to choose default port which is `465`).

Also, see the docstring in [line 26](https://github.com/cloudmarker/cloudmarker/pull/98/files#diff-28d4faa43f2d620f5a2c54123370ee47R26) for more details on this.

s-nirali: The docstring does not say that `use_ssl` can only be used for port 465. I'm not sure if someone running into this error would be able to figure out that both the `port` and `use_ssl` work together. This is not a breaking change so I'm fine if we don't do error handling here.

cloudmarker/alerts/emailalert.py
susam: It is not okay to send raw JSON in the email body. I would like `record['com']['description']` and `record['com']['recommendation']` to be picked out from the event record (`record`) and formatted neatly in the email alerts. This will be done in a separate pull request.

cloudmarker/alerts/emailalert.py
susam: [Fixed](https://github.com/cloudmarker/cloudmarker/pull/98/files#diff-28d4faa43f2d620f5a2c54123370ee47R30)

cloudmarker/alerts/emailalert.py
susam: [Fixed](https://github.com/cloudmarker/cloudmarker/pull/98/files#diff-28d4faa43f2d620f5a2c54123370ee47R23)

cloudmarker/alerts/emailalert.py
susam: I tried to address this by adding more details in the `util.send_email()` docstring in PR #108. See the following two URLs on the Read the Docs:

- [cloudmarker.alerts.emailalert.EmailAlert()](https://cloudmarker.readthedocs.io/en/latest/api/cloudmarker.alerts.html#module-cloudmarker.alerts.emailalert)
- [cloudmarker.util.send_email()](https://cloudmarker.readthedocs.io/en/latest/api/cloudmarker.html#cloudmarker.util.send_email)

Please let me know if this helps of if we need to rework on these docstrings.

PR #107, opened 2019-04-23
Add email debug param in config

sunnysharmagts:
Add smtp debug level param in config for emailalert.

### DIFF
--- a/cloudmarker/clouds/gcpcloud.py
+++ b/cloudmarker/clouds/gcpcloud.py
@@ -26,21 +26,21 @@ _log = logging.getLogger(__name__)
 class GCPCloud:
     """GCP cloud plugin."""
 
-    def __init__(self, service_account_key_path, zone):
+    def __init__(self, key_file_path, zone):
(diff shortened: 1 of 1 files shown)


cloudmarker/alerts/emailalert.py
```diff
@@ -15,7 +15,7 @@ class EmailAlert:
     """A plugin to send email notification for anomalies found."""
 
     def __init__(self, host, port, subject, to, sender, body, use_ssl=True,
-                 username=None, password=None):
+                 username=None, password=None, debug=False):
```
s-nirali: I was hoping to keep it internal like `_debug` the way it is for `azcloud` plugin. We can decide based on @susam 's opinion. Since this would only be meant for internal development and testing and not for prod use.

susam: I like `debug` because this can be used by the user too for debugging purpose.

By the way, this pull request would be overwritten by #108.

PR #108, opened 2019-04-23
Send email notifications for audits and jobs

susam:
Resolves: #53

### DIFF
--- a/cloudmarker/alerts/emailalert.py
+++ b/cloudmarker/alerts/emailalert.py
@@ -1,11 +1,9 @@
 """Email alert plugin."""
 
 
-import email
 import logging
-import smtplib
-from email.mime.multipart import MIMEMultipart
-from email.mime.text import MIMEText
+
+from cloudmarker import util
 
 # Define module-level logger.
 _log = logging.getLogger(__name__)
@@ -14,47 +12,19 @@ _log = logging.getLogger(__name__)
 class EmailAlert:
     """A plugin to send email alerts."""
 
-    def __init__(self, from_addr, to_addrs,
-                 subject='Cloudmarker Alert',
-                 host='', port=0, use_ssl=True,
-                 username=None, password=None):
+    def __init__(self, **kwargs):
         """Create an instance of :class:`EmailAlert` plugin.
 
-        When ``use_ssl`` is ``True`` and ``host`` is not specified or
-        specified as ``''``, the local host is used. When ``use_ssl`` is
-        ``True`` and ``port`` is not specified or specified as ``0``,
-        the standard SMTP-over-SSL, i.e., port 465, is used. See
-        :class:`smtplib.SMTP_SSL` documentation for more details on
-        this.
+        This class accepts the same arguments as
+        :func:`cloudmarker.util.send_email`.
 
-        When ``use_ssl`` is ``False`` and if ``host`` or ``port`` are not
-        specified, i.e., if host or port are ``''`` and/or ``0``
-        respectively, the OS default behavior is used. See
-        :class:`smtplib.SMTP` documentation for more details on this.
-
-        If ``username`` is not specified or specified as ``None``, no
-        SMTP authentication is done. If ``username`` is specified and it
-        is not ``None``, then SMTP authentication is done.
-
-        Arguments:
-            from_addr (str): Sender's email address.
-            to_addrs (list): A list of :obj:`str` objects where each
-                :obj:`str` object is a recipient's email address.
-            host (str): SMTP host.
-            port (int): SMTP port.
-            use_ssl (bool): Use SSL if ``True``, not otherwise.
-            username (str): SMTP username.
-            password (str): SMTP password.
+        The ``content`` argument is not honoured. Even if a ``content``
+        argument is provided, it is ignored by this class because this
+        class defines its own content from the event records it receives
+        in its :meth:`write` method.
         """
-        self._host = host
-        self._port = port
-        self._subject = subject
-        self._from_addr = from_addr
-        self._to_addrs = to_addrs
-        self._stringbuffer = []
-        self._use_ssl = use_ssl
-        self._username = username
-        self._password = password
+        self._kwargs = kwargs
+        self._buffer = []
 
     def write(self, record):
         """Save event record in a buffer.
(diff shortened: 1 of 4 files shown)


cloudmarker/alerts/emailalert.py
```diff
@@ -1,119 +1,40 @@
...
-            smtp = smtplib.SMTP(host=self._host, port=self._port)
-            smtp.set_debuglevel(True)
-            smtp.starttls()
-
-        if self._username:
-            smtp.login(self._username, self._password)
-        return smtp
+        """Send the buffered events as an email alert."""
+        self._kwargs['content'] = '\n\n'.join(self._stringbuffer)
+        util.send_email(**self._kwargs)
```
susam: I believe this file is easier to review if we just take a look at the final code instead of the diffs. You can do that with `git checkout auditemails; vim cloudmarker/alerts/emailalert.py` or you can just see the following code which I have included for your convenience:

```python
"""Email alert plugin."""


import logging

from cloudmarker import util

# Define module-level logger.
_log = logging.getLogger(__name__)


class EmailAlert:
    """A plugin to send email alerts."""

    def __init__(self, **kwargs):
        """Create an instance of :class:`EmailAlert` plugin.

        This class accepts the same arguments as
        :func:`cloudmarker.util.send_email`.

        The ``content`` argument is not honoured. Even if a ``content``
        argument is provided, it is ignored by this class because this
        class defines its own content from the event records it receives
        in its :meth:`write` method.
        """
        self._kwargs = kwargs
        self._buffer = []

    def write(self, record):
        """Save event record in a buffer.

        Arguments:
            record (dict): An event record.
        """
        for _, value in record.items():
            self._buffer.append(repr(value))

    def done(self):
        """Send the buffered events as an email alert."""
        self._kwargs['content'] = '\n\n'.join(self._buffer)
        util.send_email(**self._kwargs)
```

cloudmarker/util.py
```diff
@@ -453,6 +455,118 @@ def pluralize(count, word, *suffixes):
...
+            :meth:`SMTP.set_debuglevel` to debug an SMTP session. Set to
+            ``0`` (the default) or ``False`` to disable debugging. Set
+            to ``1`` or ``True`` to see SMTP messages. Set to ``2`` to
+            see timestamped SMTP messages.
+    """
+    log_data = ('from_addr: {}; to_addrs: {}; subject: {}; host: {}; '
+                'port: {}; ssl_mode: {}'
+                .format(from_addr, to_addrs, subject, host, port, ssl_mode))
+    try:
+        if ssl_mode == 'ssl':
```
sunnysharmagts: Can we define constants for these `ssl_modes` rather than hardcoding it.

cloudmarker/util.py
```diff
@@ -453,6 +455,118 @@ def pluralize(count, word, *suffixes):
...
+        smtp.send_message(msg)
+        smtp.quit()
+
+        _log.info('Sent email successfully; %s', log_data)
+
+    except Exception as e:
+        _log.error('Failed to send email; %s; error: %s: %s', log_data,
+                   type(e).__name__, e)
+
+
```
sunnysharmagts: I think it would be good to keep all the exception in one `exception` package rather than adding it to the `util.py`.  The `util.py` should only be used for generic implementations which can be used all over the project by any module.

cloudmarker/alerts/emailalert.py
```diff
@@ -1,119 +1,41 @@
...
 _log = logging.getLogger(__name__)
 
 
 class EmailAlert:
-    """A plugin to send email notification for anomalies found."""
+    """A plugin to send email alerts."""
 
-    def __init__(self, host, port, subject, to, sender, body, use_ssl=True,
-                 username=None, password=None):
+    def __init__(self, **kwargs):
```
sunnysharmagts: Since you have remove the arguments from EmailAlert plugin, it would be safe to remove `R0902`
from `pylama.ini` for emailalert.

> [pylama:cloudmarker/alerts/emailalert.py]
> ignore = R0913,C0412,R0902 //Please remove R0902

cloudmarker/util.py
susam: What would be the benefit of adding constants for these? I don't see any benefit because we are not reusing this string anywhere else. I do see a disadvantage: adding an additional level of indirection. There needs to be some justification for adding an additional level of indirection like this. Do you see any justification for it?

cloudmarker/util.py
susam: I am confused. How is the `_log.error` statement related to keeping exceptions in an `exception` package? Can you elaborate a little more about what exactly would you like me to do?

cloudmarker/util.py
sunnysharmagts: It would be easier for the plugin developers who would be using `_send_email` would have a list of `ssl_mode` options. Though the docstring gives a very crystal clear picture of the ssl mode, providing constants will give an easy options to show all the ssl_mode available.

cloudmarker/util.py
sunnysharmagts: Please refer to line no. 570. I don't know why the line is not getting reflected in the comment.
Exceptions like `PluginError`, `PortRangeError`, `PluralizeError` should be in one `exception` package

cloudmarker/alerts/emailalert.py
susam: Good point. Turns out I can remove the entire `[pylama:cloudmarker/alerts/emailalert.py]` section here. [Fixed here at line 30](https://github.com/cloudmarker/cloudmarker/pull/108/files#diff-21c09aff342003d14df85a14eeaae393L30).

cloudmarker/util.py
susam: If I understand your point correctly, the rationale here is that having the constants together at some place, say, `util.SSL`, `util.STARTTLS`, and `util.DISABLED` is going to be easier for plugin developers. We could choose another module say `constants` instead of `util` to keep the constants; that is beside the point.

I am not convinced about this point though. In fact, the opposite seems to be true to me. Let us imagine a plugin developer story for a while:

As a plugin developer, I would go to the `cloudmarker.util` documentation on [Read The Docs: cloudmarker.util](https://cloudmarker.readthedocs.io/en/latest/api/cloudmarker.html#module-cloudmarker.util). I will jump to the documentation for `send_email()`. Now I would read something like this:

> ssl_mode (str): SSL mode to use: `util.SSL` for SSL/TLS connection (the default), `util.STARTTLS` for STARTTLS, and `util.DISABLE` to disable SSL.

Now instead of the actual strings I care about, I see some constants. It gives me little idea of what the actual strings look should look like in the configuration. Granted I can just click on `util.STARTTLS` to jump to the docstring of `util.STARTTLS` and see that it is defined as `'starttls'`.

But what did we achieve with this? I don't think we achieved ease of use. It appears to me that we just increased a level of indirection, interrupted the reader, made him/her jump to another section of the documentation unnecessarily when he/she was much rather interested in reading the `util.send_email()` documentation, and increased the number of lines of code.

The current documentation reads like this:

> ssl_mode (str): SSL mode to use: `'ssl'` for SSL/TLS connection (the default), `'starttls'` for STARTTLS, and `'disable'` to disable SSL.

I find this much more developer and reader friendly. It also tells the user instantly what to put in the configuration file.

cloudmarker/util.py
susam: Could you please justify your recommendations? Why should they be in one `exception` package? What benefits does following such a practice offers?

More importantly, are those benefits so strong that they warrant breaking existing conventions? In most languages I have worked with such as C++, Java, and Python, exceptions always belong to the same namespace, package, or module, respectively. I have never seen an example where a new namespace, package, or module was created only to keep exceptions.

That makes sense too. For example, if I am parsing JSON with `json` module and I want to handle exceptions for it, I don't want to go searching the entire documentation to see what exceptions I might want to handle. I would like to type `json.` and rely on autocomplete to show me all the exceptions or just read the `json` module documentation. Similarly, we expect `smtplib` exceptions to be in the `smtplib` module. Similarly, in Java, we expect I/O exceptions to be in `java.io`, SQL exception to be in `java.sql`, and so on. In C++, for example, an I/O exception is found in `std::ios_base` namespace.

PR #129, opened 2019-05-07
Add VMDataDiskEncryptionEvent plugin

mitprasoon:
The `VMDataDiskEncryptionEvent` plugin evaluates records of type
`compute` and generates events of type `vm_data_disk_encryption_event`
one each for all the unencrypted data disks.

This plugin will generate event only if the `cloud_type` is `azure`.


cloudmarker/events/vmdatadiskencryptionevent.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+
+
+def _get_azure_vm_data_disk_encryption_event(com, ext, raw):
+    """Evaluate Azure VM for unencrypted data disks.
+
+    Arguments:
+        com (dict): Virtual Machine record `com` bucket
+        ext (dict): Virtual Machine record `ext` bucket
+        raw (dict): Virtual Machine record `raw` bucket
```
susam: In lines 76, 77, and 78: `Virtual Machine` &rarr; `Virtual machine` (normal capitalization in docstrings).

cloudmarker/events/vmdatadiskencryptionevent.py
```diff
@@ -0,0 +1,122 @@
...
+                         record)
+            return
+
+        # This plugin understands compute rule records only,
+        # so ignore  any other record types.
+        common_record_type = com.get('record_type')
+
+        if common_record_type != 'compute':
+            return
+
```
susam: I think we can put `if com.get('cloud_type') != 'azure': return` here itself. The code at line 61 below would become simpler then with one indentation level less.

cloudmarker/events/vmdatadiskencryptionevent.py
```diff
@@ -0,0 +1,122 @@
...
+    Arguments:
+        com (dict): Virtual Machine record `com` bucket
+        ext (dict): Virtual Machine record `ext` bucket
+        raw (dict): Virtual Machine record `raw` bucket
+    Returns:
+        dict: An event record representing unencrypted data disk
+
+    """
+    friendly_cloud_type = util.friendly_string(com.get('cloud_type'))
+    os_disk_name = raw.get('storage_profile').get('os_disk').get('name')
```
susam: This needs to become:

```
raw.get('storage_profile', {}).get('os_disk', {}).get('name')
```

so that if `raw.get('storage_profile')` returns `None`, the code still wouldn't crash.

cloudmarker/events/vmdatadiskencryptionevent.py
```diff
@@ -0,0 +1,122 @@
...
+    instance_view = raw.get('instance_view')
+    if instance_view is None:
+        return
+    for disk in instance_view.get('disks'):
+        # If it is an OS disk skip and continue
+        if disk.get('name') == os_disk_name:
+            continue
```
susam: Do you plan to handle OS disk encryption here itself or do you plan to write a plugin for this? I am thinking we could handle OS disk encryption here itself. Of course, it does not have to be done in this pull request but I just wanted to know your thoughts about it.

cloudmarker/events/vmdatadiskencryptionevent.py
```diff
@@ -0,0 +1,122 @@
...
+        if disk.get('encryption_settings') is not None and \
+                disk['encryption_settings'][0]['enabled']:
```
susam: Instead of checking only `disk['encryption_settings'][0]` here, we might need to loop over all items in `disk['encryption_settings']` and then check each item's `'enabled'` key. I am only guessing here. You are the expert in this area, so you would be able to take the best decision here.

If you also think we need to loop here, it might be useful to create a separate function to do the looping work to prevent this part of the code from becoming more deeply nested.

cloudmarker/test/test_vmdatadiskencryptionevent.py
```diff
@@ -0,0 +1,153 @@
...
+import unittest
+
+from cloudmarker.events import vmdatadiskencryptionevent
+
+base_record = {
+    'ext':  {
+        'cloud_type':  'azure',
+        'record_type':  'virtual_machine',
+        'subscription_id':  'sub_id_0000',
+        'all_data_disks_encrypted':  True
```
susam: I don't see the tests depending on `cloud_type`, `record_type`, and `subscription_id`, so I think these fields can be removed from `base_record`. The unit tests should be set up with as minimal data as possible, so that the tests are very focused and reading the tests also easily shows us what it is being that is tested.

cloudmarker/test/test_vmdatadiskencryptionevent.py
```diff
@@ -0,0 +1,153 @@
...
+    },
+    'com':  {
+        'cloud_type':  'azure',
+        'record_type':  'compute',
+        'reference':  'reference'
```
susam: I don't see any test depending on the `reference` field, so I think this field can also be removed.

cloudmarker/test/test_vmdatadiskencryptionevent.py
```diff
@@ -0,0 +1,153 @@
...
+        record['raw']['instance_view']['disks'].append(encrypted_data_disk)
+        plugin = vmdatadiskencryptionevent.VMDataDiskEncryptionEvent()
+        events = list(plugin.eval(record))
+        self.assertEqual(len(events), 0)
+
+    def test_multiple_unencrypted_data_disk(self):
+        record = copy.deepcopy(base_record)
+        record['raw'] = copy.deepcopy(base_raw_bucket)
+        record['ext']['all_data_disks_encrypted'] = False
+        unencrypted_data_disks = random.randint(1, 5)
```
susam: Why not have a definite number of disks here like say `3` or `5` instead of randomizing the number of disks?

cloudmarker/events/vmdatadiskencryptionevent.py
```diff
@@ -0,0 +1,122 @@
...
+_log = logging.getLogger(__name__)
+
+
+class VMDataDiskEncryptionEvent:
+    """VM Data disk encryption event plugin."""
+
+    def __init__(self):
+        """Create an instance of :class:`VMDataDiskEncryptionEvent`."""
+
+    # pylint: disable=R0201
```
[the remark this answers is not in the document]
mitprasoon: Removed pylint ignore directive to `pylama.ini`

cloudmarker/events/vmdatadiskencryptionevent.py
```diff
@@ -0,0 +1,122 @@
...
+        # If it is an OS disk skip and continue
+        if disk.get('name') == os_disk_name:
+            continue
+        if disk.get('encryption_settings') is not None and \
+                disk['encryption_settings'][0]['enabled']:
+            continue
+        reference = com.get('reference')
+        description = (
+            '{} virtual machine {} has unencrypted Data disk {}'
+            .format(friendly_cloud_type, reference, disk.get('name'))
```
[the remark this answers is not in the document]
mitprasoon: Changed 'Data' to 'data' in the message

cloudmarker/events/vmdatadiskencryptionevent.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+import logging
+
+from cloudmarker import util
+
+_log = logging.getLogger(__name__)
+
+
+class VMDataDiskEncryptionEvent:
```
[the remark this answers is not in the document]
mitprasoon: Renamed plugin to `AzVMDataDiskEncryptionEvent
`Az` suggests that this plugin is for Azure cloud while `VMDataDisk` suggests that this plugin is for data disks attached to virtual machine

cloudmarker/events/vmdatadiskencryptionevent.py
mitprasoon: In line lines 76, 77, and 78:  changed `Virtual Machine` to `Virtual machine`

cloudmarker/events/vmdatadiskencryptionevent.py
mitprasoon: Having  `if com.get('cloud_type') != 'azure': return` results in the following issue:
`E701 multiple statements on one line (colon) [pycodestyle]`
Suggestion?

cloudmarker/events/vmdatadiskencryptionevent.py
mitprasoon: I plan to have another plugin for OS disk encryption event for the following reasons:
1. OS disk encryption is a prerequisite for data disk encryption
2. Having a different plugin for OS  disk encryption will provide a more granular control.

cloudmarker/events/vmdatadiskencryptionevent.py
mitprasoon: Having tried multiple scenarios I haven't seen a case where the size of `disk['encryption_settings']` is more than 1. I plan to keep it as it, till the time it breaks.

cloudmarker/test/test_vmdatadiskencryptionevent.py
mitprasoon: Removed `cloud_type`, `record_type` and `subscription_id`

cloudmarker/test/test_vmdatadiskencryptionevent.py
mitprasoon: Removed `reference`

cloudmarker/test/test_vmdatadiskencryptionevent.py
mitprasoon: My intention was to introduce randomness to the test case. Suggestions?

PR #130, opened 2019-05-08
Add AzVMOSDiskEncryptionEvent plugin

mitprasoon:
The `AzVMOSDiskEncryptionEvent` plugin evaluates records of type
`compute` and generates events of type `vm_os_disk_encryption_event`
for unencrypted virtual machine OS disk.

This plugin will generate event only if the `cloud_type` is `azure`.

### DIFF
--- a/cloudmarker/events/azvmosdiskencryptionevent.py
+++ b/cloudmarker/events/azvmosdiskencryptionevent.py
@@ -1,0 +1,110 @@
+"""Microsoft Azure VM OS disk encryption event.
+
+This module defines the :class:`AzVMOSDiskEncryptionEvent` class that
+identifies an unencrypted Azure OS disk. This plugin works on the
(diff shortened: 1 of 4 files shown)


cloudmarker/events/azvmosdiskencryptionevent.py
```diff
@@ -0,0 +1,110 @@
...
+        # This plugin understands compute rule records only,
+        # so ignore  any other record types.
+        common_record_type = com.get('record_type')
+
+        if common_record_type != 'compute':
+            return
+
+        # If 'ext' bucket is missing, we have a malformed record. Log a
+        # warning and ignore it.
+        ext = record.get('ext')
```
sunnysharmagts: Can we put these checks in a single method which logs warning in case a specific bucket is missing.

PR #140, opened 2019-05-22
Add PostgresSQL to resources evaluated by AzCloud

mitprasoon:
The PostgreSQL servers associated with a subscription are evaluated
and processed. A record is generated, with the value of `record_type`
set to `postgresql_server` in the `ext` bucket and `record_type`
set to `rdbms` set in the `com` bucket of the record.

### DIFF
--- a/cloudmarker/test/test_azcloud.py
+++ b/cloudmarker/test/test_azcloud.py
@@ -72,6 +72,10 @@ class AzCloudTest(unittest.TestCase):
         self._MockMySQLManagementClient = m
         m().servers.list.return_value = [mock_record]
 
+        m = self._patch('PostgreSQLManagementClient')
+        self._MockPostgreSQLManagementClient = m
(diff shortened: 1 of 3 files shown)


cloudmarker/test/test_azcloud.py
```diff
@@ -580,3 +584,27 @@ def test_mysql_server_record(self):
...
+
+        m = self._MockPostgreSQLManagementClient
+        m().servers.list.return_value = [mock_postgresql_server]
+
+        records = list(azcloud.AzCloud('', '', '').read())
+        records = [
+            r for r in records
+            if r['ext']['record_type'] == 'postgresql_server'
+        ]
+        self.assertEqual(records[0]['ext']['ssl_connection_enabled'],
```
susam: I believe this assert would change to `self.assertEqual(records[0]['com']['tls_enforced'], True)` after this pull request is rebased on the latest `master`.

PR #141, opened 2019-05-22
Add RDBMSEnforceTLSEvent plugin

mitprasoon:
The `RDBMSEnforceTLSEvent` plugin evaluates the records of type
'rdbms' and generates events of type `rdbms_enforce_tls_event` if
TLS/SSL enforcement is disabled on the server.

Enforcing TLS/SSL connection between the server and the client applications
help prevent `Man In The Middle ` attack by encrypting data stream
between the server and the client application.

### DIFF
--- a/cloudmarker/events/rdbmsenforcetlsevent.py
+++ b/cloudmarker/events/rdbmsenforcetlsevent.py
@@ -1,0 +1,100 @@
+"""RDBMS Enforce TLS/SSL Event.
+
+This module defines the :class:`RDBMSEnforceTLSEvent` class that
+identifies RDBMS servers which have TLS/SSL connection enforcement
+disabled. This plugin works on the  properties found in the ``com``
+bucket of ``rdbms`` records.
+"""
+
+
+import logging
+
+from cloudmarker import util
+
+_log = logging.getLogger(__name__)
+
+
+class RDBMSEnforceTLSEvent:
+    """Az RDBMS TLS/SSL enforcement event plugin."""
+
+    def __init__(self):
+        """Create an instance of :class:`RDBMSEnforceTLSEvent`."""
+
+    def eval(self, record):
+        """Evaluate RDBMS servers for TLS connection enforcement.
+
+        Arguments:
+            record (dict): An RDBMS record.
+
+        Yields:
+            dict: An event record representing an RDBMS where TLS
+            connection enforcement is disabled
+
+        """
+        com = record.get('com', {})
+        if com is None:
+            return
+
+        if com.get('record_type') != 'rdbms':
+            return
+
+        ext = record.get('ext', {})
+        if ext is None:
+            return
+
+        # True, None, missing key or any other value will not
+        # genarated an event. An event will be generated only if
+        # the value of `tls_enforced` is False.
+        if com.get('tls_enforced') is False:
+            yield from _get_rdbms_tls_enforcement_event(
+                com, ext)
+
+    def done(self):
+        """Perform cleanup work.
+
+        Currently, this method does nothing. This may change in future.
+        """
+
+
+def _get_rdbms_tls_enforcement_event(com, ext):
+    """Generate event for TLS enforcement disabled.
+
+    Arguments:
+        com (dict): RDBMS record `com` bucket
+        ext (dict): RDBMS record `ext` bucket
+    Returns:
+        dict: An event record representing RDBMS with SSL
+        connection enforcement disabled
+
+    """
+    friendly_cloud_type = util.friendly_string(com.get('cloud_type'))
+    friendly_rdbms_type = util.friendly_string(ext.get('record_type'))
+
+    reference = com.get('reference')
+    description = (
+        '{} {} {} has TLS/SSL enforcement disabled.'
+        .format(friendly_cloud_type, friendly_rdbms_type, reference)
+    )
+    recommendation = (
+        'Check {} {} {} and enable TLS/SSL enforcement.'
+        .format(friendly_cloud_type, friendly_rdbms_type, reference)
+    )
+
+    event_record = {
+        # Preserve the extended properties from the RDBMS
+        # record because they provide useful context to
+        # locate the RDBMS that led to the event.
+        'ext': util.merge_dicts(ext, {
+            'record_type': 'rdbms_enforce_tls_event'
+        }),
+        'com': {
+            'cloud_type': com.get('cloud_type'),
+            'record_type': 'rdbms_enforce_tls_event',
+            'reference': reference,
+            'description': description,
+            'recommendation': recommendation,
+        }
+    }
+
+    _log.info('Generating rdbms_enforce_tls_event; %r', event_record)
+    yield event_record
--- a/cloudmarker/clouds/azcloud.py
+++ b/cloudmarker/clouds/azcloud.py
@@ -472,10 +472,10 @@ def _get_normalized_rdbms_record(rdbms_record):
         'raw': rdbms_record.get('raw', {}),
         'ext': util.merge_dicts(rdbms_record.get('ext'), {
             'reference': rdbms_record.get('raw', {}).get('id'),
-            'ssl_connection_enabled': ssl_connection_enabled,
         }),
         'com': util.merge_dicts(rdbms_record.get('com'), {
             'reference': rdbms_record.get('raw', {}).get('id'),
+            'tls_enforced': ssl_connection_enabled,
         }),
 
     }
--- a/cloudmarker/test/test_azcloud.py
+++ b/cloudmarker/test/test_azcloud.py
@@ -572,7 +572,7 @@ class AzCloudTest(unittest.TestCase):
             r for r in records
             if r['ext']['record_type'] == 'mysql_server'
         ]
-        self.assertEqual(records[0]['ext']['ssl_connection_enabled'],
+        self.assertEqual(records[0]['com']['tls_enforced'],
                          True)
         self.assertEqual(records[0]['ext']['reference'],
                          'azure_mysql_server_id')
--- a/cloudmarker/test/test_rdbmsenforcetlsevent.py
+++ b/cloudmarker/test/test_rdbmsenforcetlsevent.py
@@ -1,0 +1,63 @@
+"""Tests for RDBMSEnforceTLSEvent plugin."""
+
+
+import copy
+import unittest
+
+from cloudmarker.events import rdbmsenforcetlsevent
+
+base_record = {
+    'com':  {
+        'tls_enforced':  True,
+        'record_type': 'rdbms',
+    }
+}
+
+
+class RDBMSEnforceTLSEventTest(unittest.TestCase):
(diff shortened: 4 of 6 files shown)


cloudmarker/events/azrdbmssslenforcementevent.py
```diff
@@ -0,0 +1,104 @@
...
+
+import logging
+
+from cloudmarker import util
+
+_log = logging.getLogger(__name__)
+
+
+class AzRDBMSSSLEnforcementEvent:
+    """Az RDBMS SSL enforcement event plugin."""
```
susam: I like one of your suggestions on chat about renaming this plugin to `AzRDBMSEnforceSSLEvent`.

In fact, I also think that this can be a generic multi-cloud plugin that works for both Azure and GCP clouds. Whether TLS connectivity is enforced or not is a generic feature of all DBs in all clouds. There are only three possible answers: yes, no, or unknown which we can capture as `True`, `False`, or `None`. In that case the name `RDBMSEnforceSSL` or `RDBMSEnforceTLS` would be more suitable.

cloudmarker/events/azrdbmssslenforcementevent.py
```diff
@@ -0,0 +1,104 @@
...
+
+        ext = record.get('ext', {})
+        if ext is None:
+            return
+
+        if ext.get('record_type') not in ['mysql_server',
+                                          'postgresql_server']:
+            return
+
+        if ext.get('ssl_connection_enabled', False) is False:
```
susam: The idiomatic way to write this is:

```
if not ext.get('ssl_connection_enabled'):
```

cloudmarker/events/azrdbmssslenforcementevent.py
```diff
@@ -0,0 +1,104 @@
...
+            return
+
+        if com.get('cloud_type') != 'azure':
+            return
+
+        ext = record.get('ext', {})
+        if ext is None:
+            return
+
+        if ext.get('record_type') not in ['mysql_server',
```
susam: If you decide to make this a generic plugin, then this check would change to `com.get('record_type') != 'rdbms'`.

cloudmarker/events/azrdbmssslenforcementevent.py
susam: I think the field name `ssl_connection_enabled` should be changed to `ssl_enforced`. I am trying to keep the names as simple and short as possible. SSL is always meant for connections, so the "connection" word is not necessary. For example, Microsoft has named this field as `ssl_enforcement` in their API response, so `ssl_enforced` is good for us.

Note: If this is going to be a generic plugin then we can go for a vendor neutral term and choose `tls_enforced` as the name. "TLS" is the modern name and spec for this protocol anyway, so I think it makes a good name.

cloudmarker/events/azrdbmssslenforcementevent.py
```diff
@@ -0,0 +1,104 @@
...
+    friendly_rdbms_type = util.friendly_string(ext.get('record_type'))
+
+    reference = com.get('reference')
+    description = (
+        '{} {} {} has SSL connection enforcement disabled'
+        .format(friendly_cloud_type, friendly_rdbms_type, reference)
+    )
+    recommendation = (
+        'Check {} {} {} and enable SSL connection enforcement'
+        .format(friendly_cloud_type, friendly_rdbms_type, reference)
```
susam: Please end `recommendation` and `description` messages with a full-stop. The intention here is to write human-friendly complete sentences that could be picked up from the event and used directly in emails if necessary.

I see that the same update is required in the disk encryption event plugins. A separate pull request can be created for those plugins.

cloudmarker/events/azrdbmssslenforcementevent.py
mitprasoon: Renamed the plugin to `RDBMSEnforceTLSEvent`

cloudmarker/events/azrdbmssslenforcementevent.py
mitprasoon: I want to make generate an event only if the value of `ssl_connection_enabled ` is False.
`if not ext.get('ssl_connection_enabled'):
 `
may evaluate to True if the value of `ssl_connection_enabled`  is anything other than `False` e.g. if `ssl_connection_enabled` is `None` the above condition will be True

cloudmarker/events/azrdbmssslenforcementevent.py
mitprasoon: Removed the logic which checks for 'record_type' in the `ext` bucket and matches it against certain RDBMS types.

cloudmarker/events/azrdbmssslenforcementevent.py
mitprasoon: In an effort to make this plugin cloud agnostic `ssl_connection_enabled ` has been changed to `tls_enforced`

cloudmarker/events/azrdbmssslenforcementevent.py
mitprasoon: Corrected  `recommendation` and `description` so that they end with a full-stop.

cloudmarker/events/azrdbmssslenforcementevent.py
susam: You are right. I wrote it wrong. I meant to say,

```
if not ext.get('ssl_connection_enabled', False):
```

Does this make sense?

cloudmarker/events/rdbmsenforcetlsevent.py
```diff
@@ -0,0 +1,97 @@
...
+            return
+
+        if com.get('record_type') != 'rdbms':
+            return
+
+        ext = record.get('ext', {})
+        if ext is None:
+            return
+
+        if com.get('tls_enforced', True) is False:
```
susam: This is equivalent to the idiomatic form:

```
if not com.get('tls_enforced', True):
```

cloudmarker/events/azrdbmssslenforcementevent.py
mitprasoon: `if not ext.get('ssl_connection_enabled', False):` evaluates to `True` if the value of `ssl_connection_enabled` is anything other than `False` (e.g. `None`) or the key `ssl_connection_enabled` is missing from the `ext` bucket

susam: You are right again. I am getting confused with the case of `None` as a value. My

cloudmarker/events/rdbmsenforcetlsevent.py
susam: I wrote a simple program and found out that what I am suggesting is indeed not equivalent to what you have written:

```python
# Solution 1.
com = {}
for flag_value in -1, True, None, False:
    com['tls_enforced'] = flag_value
    if flag_value == -1: # Missing key case.
        del com['tls_enforced']

    # Your correct solution.
    if com.get('tls_enforced', True) is False:
        print('bad')
    else:
        print('good')
print()


# Solution 2
com = {}
for flag_value in -1, True, None, False:
    com['tls_enforced'] = flag_value
    if flag_value == -1: # Missing key case.
        del com['tls_enforced']

    # My incorrect solution.
    if not com.get('tls_enforced', True):
        print('bad')
    else:
        print('good')
print()
```

Output:

```
good
good
good
bad

good
good
bad
bad
```

Having understood this, I still find specifying a default value of `True` in `com.get()` redundant. For example, see this code:

```python
# Solution 3
com = {}
for flag_value in -1, True, None, False:
    com['tls_enforced'] = flag_value
    if flag_value == -1: # Missing key case.
        del com['tls_enforced']

    # My suggestion: No default in com.get().
    if com.get('tls_enforced') is False:
        print('bad')
    else:
        print('good')
print()
```

Output:

```
good
good
good
bad
```

If the key is missing, `com.get()` would default to `None` anyway and you are already handing `None` as a good thing by the strict comparison with `is False`.

Since this part of the code took a while for me to understand, I think a comment would be very helpful for an occassionally dim-witted reader like me. 🙂 For example:

```python
    # True, None, or missing key is considered good. Only False is
    # considered a violation and leads to an event.
    if not com.get('tls_enforced') is False:
```

mitprasoon: Indeed, having a default value adds no value. I have changed the condition to: 
` if com.get('tls_enforced') is False:
`
and have added appropriate comments.

cloudmarker/events/rdbmsenforcetlsevent.py
```diff
@@ -0,0 +1,100 @@
...
+
+        if com.get('record_type') != 'rdbms':
+            return
+
+        ext = record.get('ext', {})
+        if ext is None:
+            return
+
+        # True, None, missing key or any other value will not
+        # genarate an event. An event will be generate only if
```
susam: Typo: "will be generate" => "will be generated"

cloudmarker/events/rdbmsenforcetlsevent.py
```diff
@@ -0,0 +1,100 @@
...
+
+def _get_rdbms_tls_enforcement_event(com, ext):
+    """Generate event for TLS enforcement disabled.
+
+    Arguments:
+        com (dict): RDBMS record `com` bucket
+        ext (dict): RDBMS record `ext` bucket
+    Returns:
+        dict: An event record representing RDBMS with SSL
+        connection enforcement disabled
```
susam: The second line, i.e., "connection enforcement disabled", needs to be indented one level more. Sorry, forgot to spot this in an earlier review.

cloudmarker/events/rdbmsenforcetlsevent.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def eval(self, record):
+        """Evaluate RDBMS servers for TLS connection enforcement.
+
+        Arguments:
+            record (dict): A RDBMS record.
+
+        Yields:
+            dict: An event record representing an RDBMS where TLS
+            connection enforcement is disabled
```
susam: The second line "connection enforcement is disabled" needs to be indented one level more.

cloudmarker/events/rdbmsenforcetlsevent.py
mitprasoon: Fixed the type

cloudmarker/events/rdbmsenforcetlsevent.py
mitprasoon: From Google Python Style Guide section [doc function returns](https://google.github.io/styleguide/pyguide.html#doc-function-returns) no indentation is needed.
Thanks @susam  for the explanation.

cloudmarker/events/rdbmsenforcetlsevent.py
mitprasoon: From Google Python Style Guide section [doc function returns](https://google.github.io/styleguide/pyguide.html#doc-function-returns) no indentation is needed.
Thanks @susam  for the explanation.

PR #144, opened 2019-06-07
Remove zone parameter in GCPCloud

jaibhageria:
Using the `zones().list()` API provided by google-api-python-client
library, the available zones for a given project name can be obtained.
Each zone is then checked for virtual machine instances and the details
along with relevant firewall rules can be recorded.

### DIFF
--- a/cloudmarker/clouds/gcpcloud.py
+++ b/cloudmarker/clouds/gcpcloud.py
@@ -26,16 +26,14 @@ _log = logging.getLogger(__name__)
 class GCPCloud:
     """GCP cloud plugin."""
 
-    def __init__(self, key_file_path, zone):
+    def __init__(self, key_file_path):
         """Create an instance of :class:`GCPCloud` plugin.
 
         Arguments:
             key_file_path (str): Path of the service account
                 key file for a project.
-            zone (str): Zone of GCP Project, e.g., ``us-east1-b``.
         """
         self._key_file_path = key_file_path
-        self._zone = zone
 
         # Service account key file also has the project name under the key
         # project_id. We will use this key file to get the project name for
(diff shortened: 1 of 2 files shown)


cloudmarker/clouds/gcpcloud.py
```diff
@@ -58,6 +56,16 @@ def __init__(self, key_file_path, zone):
         # https://developers.google.com/api-client-library/python/apis/
         self._compute_resource = self._get_resource('compute', 'v1')
 
+        # Fetching all available zones for the current project.
+        get_zones = self._compute_resource.zones()
+        list_zones = get_zones.list(project=self._project_name)
+        project_zones = list_zones.execute()
```
susam: The zone iteration code should move to `read()` method. The `__init__()` method should be fairly lightweight.

cloudmarker/clouds/gcpcloud.py
```diff
@@ -121,32 +119,47 @@ def _get_instances(self):
...
-                break
-
-            next_page_token = instances_object['nextPageToken']
+        # Iterating through all zones in the zones list, and collecting
+        # virtual machine instance details for each zone.
+        for zone in zones:
+            while True:
+                # Prepare the request to get the list of all VM instances for a
+                # project and execute that request.
+                instances_object = instance_resource.list(
```
susam: Can you rename this variable to something simple, say, `instance`? Like I mentioned earlier, everything is an object anyway, so having `object` in the name is redundant.

jaibhageria: This variable name was already a part of the code, and was not introduced by me.

susam: Sounds good. I will go ahead and test this pull request and if everything looks good, merge it to `master`.

PR #145, opened 2019-06-11
Fix UserWarning in MongoDBStore

r0hi7:
Since Cloudmarker runs in multiprocess mode the
connect call in `cloudmarker/stores/mongodbstore.py`
MongoClient was throwing a `UserWarning: MongoClient
opened before fork. Create MongoClient only after
forking. See PyMongo's documentation for details:
http://api.mongodb.org/python/current/faq.html#is-pymongo-fork-safe
"MongoClient opened before fork. Create MongoClient
only "`.
In this commit the call to MongoDB via is
made lazy. The connection to the database is made on
the first operation.

### DIFF
--- a/cloudmarker/stores/mongodbstore.py
+++ b/cloudmarker/stores/mongodbstore.py
@@ -34,7 +34,8 @@ class MongoDBStore:
             host=host,
             port=port,
             username=username,
-            password=password
+            password=password,
+            connect=False
         )
 
         self._collection = self._client[db][collection]


susam:
I believe a good way to solve this problem would be to fix the underlying issue in the framework code.

Currently, the plugins are instantiated in `manager.py` and the instantiated plugins are then passed to the `multiprocessing` processes via a `multiprocessing` queue. This causes the `MongoClient` warning. In future, this could also cause other issues such as serialization error when we instantiate a plugin and send it to worker processes over the queue if the plugin object cannot be serialized. This looks like a mistake in the framework code that I have written. I would like to fix this issue in the framework code itself.

PR #146, opened 2019-06-11
Add baseconfig key for RDBMSEnforceTLSEvent

susam:
The `baseconfig.py` module contains built-in config keys that can be
referenced directly in the user's config file. In other words, a user
can use these config keys without having to define them.

This change adds an `rdbmsenforcetlsevent` config key to configure the
`RDBMSEnforceTLSEvent` plugin. This plugin is a good candidate for a
built-in base config key because this plugin accepts no parameters, so
once configured in hte base config, it can be used in any user config.


susam:
> There is no description for the PR and commit. Though it is trivial and pretty self explanatory, shall we still add it?

Hi Prasoon (@mitprasoon),

Thank you for the review comment. I have added a description both in the commit body and this pull request.

PR #147, opened 2019-06-11
Instantiate plugins in worker processes

susam:
Prior to this change, the plugin classes are instantiated in the main
process running `manager.py` and the instantiated plugins are passed to
the respective worker processes implemented in `workers.py` via
`multiprocessing.Queue` objects.

Instantiating a plugin in the main process in `manager.py` and then
passing it to a worker process via a queue is problematic. If a plugin
is not fork-safe, it may not function correctly when used inside a
worker process. One such problem was found with the `MongoDBStore`
plugin which is not fork-safe because it in turn instantiates
`MongoClient` which is not fork-safe. The following warning is
encountered while using `MongoDBStore`:

    UserWarning: MongoClient opened before fork. Create MongoClient only
    after forking. See PyMongo's documentation for details:
    http://api.mongodb.org/python/current/faq.html#is-pymongo-fork-safe
    "MongoClient opened before fork. Create MongoClient only "

There is another possible problem too. If a plugin is not picklable
(serialiazble), then attempting to pass an instantiated plugin over a
`multiprocessing.Queue` object would lead to `PickleError`.

To prevent these issues, this change implements a new strategy for
instantiating the plugins. The main process running `manager.py` passes
the plugin config dictionaries to the worker processes over
`multiprocessing.Queue` objects. The worker processes instantiate the
plugins from the plugin config dictionaries. This ensures that we pass
only a serializable object (i.e., a dictionary, not an instantiated
plugin) over the queue. It also ensures that a plugin is instantiated in
the same process where it is used, i.e., within a worker process.

Resolves: #145

### DIFF
--- a/cloudmarker/manager.py
+++ b/cloudmarker/manager.py
@@ -142,7 +142,7 @@ class Audit:
                 audit_key,
                 audit_version,
                 plugin_key,
-                util.load_plugin(config['plugins'][plugin_key]),
+                config['plugins'][plugin_key],
                 input_queue,
             )
             worker = mp.Process(target=workers.alert_worker, args=args)
@@ -156,7 +156,7 @@ class Audit:
                 audit_key,
                 audit_version,
                 plugin_key,
-                util.load_plugin(config['plugins'][plugin_key]),
+                config['plugins'][plugin_key],
                 input_queue,
                 self._alert_queues,
             )
@@ -171,7 +171,7 @@ class Audit:
                 audit_key,
                 audit_version,
                 plugin_key,
-                util.load_plugin(config['plugins'][plugin_key]),
+                config['plugins'][plugin_key],
                 input_queue,
             )
             worker = mp.Process(target=workers.store_worker, args=args)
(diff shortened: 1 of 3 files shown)


susam:
#### Reproduce the Issue

 1. Create a config file named `test.yaml`.

    ```yaml
    audits:
      mockaudit:
        stores:
          - mongodbstore
    ```

 2. Check out master branch. It does not have the fix at this time, so it can be used to reproduce the issue.

    ```
    git checkout master
    ```

 3. Run MongoDB.

    ```
    docker rm mongo; docker run --name mongo -p 27017:27017 mongo
    ```

 4. Run Cloudmarker.

    ```
    python3 -m cloudmarker -n -c test.yaml
    ```

 5. Observe the following warning in the output.

    ```
    venv/lib/python3.7/site-packages/pymongo/topology.py:150:
    UserWarning: MongoClient opened before fork. Create MongoClient only
    after forking. See PyMongo's documentation for details:
    http://api.mongodb.org/python/current/faq.html#is-pymongo-fork-safe
    "MongoClient opened before fork. Create MongoClient only "
    ```

#### Verify the Fix

 1. Check out this pull request branch. It contains the fix.

    ```
    git checkout pluginframework
    ```

 2. Run Cloudmarker again.

    ```
    python3 -m cloudmarker -n -c test.yaml
    ```

 3. The `UserWarning` message should no longer appear.

PR #148, opened 2019-06-13
Pull all projects for a keyfile in GCPCloud

jaibhageria:
Using the `projects().list()` API provided by google-api-python-client
library, all available projects for a given keyfile can be obtained.
Successively, all projects are iterated through and the instance and
firewall data is recorded.

### DIFF
--- a/cloudmarker/clouds/gcpcloud.py
+++ b/cloudmarker/clouds/gcpcloud.py
@@ -5,7 +5,6 @@ Google Cloud Platform.
 """
 
 
-import json
 import logging
 
 from google.oauth2 import service_account
@@ -13,7 +12,8 @@ from googleapiclient import discovery
 
 from cloudmarker import util
 
-_GCP_SCOPES = ['https://www.googleapis.com/auth/compute.readonly']
+_GCP_SCOPES = ['https://www.googleapis.com/auth/compute.readonly',
+               'https://www.googleapis.com/auth/cloud-platform.read-only']
 """OAuth 2.0 scopes for Google APIs required by this plugin.
 
 See https://developers.google.com/identity/protocols/googlescopes for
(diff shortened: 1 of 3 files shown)


cloudmarker/clouds/gcpcloud.py
```diff
@@ -26,21 +26,22 @@
...
     """GCP cloud plugin."""
 
-    def __init__(self, key_file_path):
+    def __init__(self, key_file_path, _max_projects=0):
         """Create an instance of :class:`GCPCloud` plugin.
 
+         Note: The ``_max_projects`` argument should be
+         used only in the development-test-debug phase. It should not
+         be used in production environment. This is why we use the
+         convention of beginning it's name with underscore.
```
susam: Each line should try to occupy as much of the 72 character soft-limit as possible without exceeding 72 characters. The hard-limit is 79 characters. In this case, the first line is much shorter (55 characters only when it could be longer).

The wrapping of lines should not be done manually. It is quite hard to get this right manually, especially, when one wants to update the lines again after formatting. You should try to figure out if your editor can help you do this. With Vim I enter `:set textwidth=72` (it's set in my `~/.vimrc`) and then use the normal mode commands `gqap` or `gwap` to format the current paragraph or line correctly into multiple lines. If you are using another editor, you would have to figure out a way to automate this, perhaps with additional plugins.

cloudmarker/clouds/gcpcloud.py
```diff
@@ -75,9 +78,13 @@ def _get_resource(self, service_name, version='v1'):
                                credentials=self._credentials,
                                cache_discovery=False)
 
-    def _get_firewalls(self):
+    def _get_firewalls(self, projectId):
```
susam: The parameter name should be `project_id`. We don't use camelCase for variable or parameter names. Please make this change at other places in the code too.

cloudmarker/clouds/gcpcloud.py
```diff
@@ -93,10 +100,13 @@ def _get_firewalls(self):
...
             firewall_rules_object = firewall_resource.list(
-                project=self._project_name,
+                project=projectId,
                 pageToken=next_page_token).execute()
 
-            firewall_rules.extend(firewall_rules_object['items'])
+            # firewall_rules_object contains firewall rules for the project
+            # if it has the 'items' key.
+            if 'items' in firewall_rules_object.keys():
+                firewall_rules.extend(firewall_rules_object['items'])
```
susam: The double-lookup of `'items'` in `firewall_rules_object` can be avoided with something like this:

```python
firewall_rules.extend(firewall_rules_object.get('items', []))
```

cloudmarker/test/test_gcpcloud.py
```diff
@@ -145,6 +194,7 @@ def test_firewall_single_allowed_rule(self):
         }
         m = self._mock_discovery
         m.build().firewalls().list().execute.return_value = mock_firewall_dict
+        m.build().projects().list().execute.return_value = mock_projects_dict
```
susam: In each of these test methods, can you put the `m.build().projects()` statements before `m.build().firewalls()`? In places like this, I like to see the mocks being setup in the chronological order of usage. In the `GCPCloud` code, the projects would be retrieved before the firewalls, so I prefer maintaining the same order in the test code as well. This is not a strict rule. There are cases where test code neatness may trump chronological ordering as per usage.

PR #149, opened 2019-06-17
Use concurrency in GCPCloud

jaibhageria:


cloudmarker/clouds/gcpcloud.py
```diff
@@ -37,29 +39,39 @@ def __init__(self, key_file_path, _max_projects=0):
...
-        # service_name and version information, the more complete information
-        # about available_service name and their corresponding version is
-        # located at
-        # https://developers.google.com/api-client-library/python/apis/
-        self._compute_resource = self._get_resource('compute', 'v1')
+        # Service account key file also has the client email under the key
+        # client_email. We will use this key file to get the client email for
+        # this request.
+        with open(self._key_file_path) as f:
+            self._client_email = json.loads(f.read())['client_email']
```
susam: Can you change this to `json.loads(f.read()).get('client_email')`, so that even if `'client_email'` key is missing, this line of code would not crash the plugin?

cloudmarker/clouds/gcpcloud.py
```diff
@@ -71,213 +88,186 @@ def _get_resource(self, service_name, version='v1'):
...
-            # 500, if the result contains more than 500 records, nextPageToken
-            # is present in response whose value is to be used to make the
-            # subsequent list() call.
-            if 'nextPageToken' not in firewall_rules_object.keys():
-                break
+                yield ('firewall', project_index, project)
 
-            next_page_token = firewall_rules_object['nextPageToken']
+                zones = _get_resource_iterator(compute_resource.zones(),
+                                               'items', project=project_id)
```
susam: While testing, I found a problem here. If the zone iteration fails for any reason, then the `for project_index, project in enumerate(projects)` loop crashes before it can pull the remaining projects. A fix for this should go into the `_get_resource_iterator()` function.

cloudmarker/clouds/gcpcloud.py
```diff
@@ -71,213 +88,186 @@ def _get_resource(self, service_name, version='v1'):
...
 
-def _get_normalized_firewall_rules(firewall_record):
+    """
+    request = resource.list(**list_kwargs)
+    while request is not None:
+        response = request.execute()
+        for item in response.get(key, []):
+            yield item
+        request = resource.list_next(previous_request=request,
+                                     previous_response=response)
```
susam: The complete content of this function should be enclosed in a `try-except` block, so that any zone discovery failure, does not crash the complete project discovery loop. Use this format for logging the exception:

```
            _log.error('Failed to fetch resource list; key: %s; '
                       'list_kwargs: %s; key_file_path: %s; '
                       'error: %s: %s', key, list_kwargs,
                       self._key_file_path, type(e).__name__, e)
```

PR #150, opened 2019-06-24
Add AzSQL plugin

mitprasoon:
The plugin `AzSQL` defines the methods to pull SQL database information
from the Microsoft Azure cloud.

### DIFF
--- a/cloudmarker/clouds/azsql.py
+++ b/cloudmarker/clouds/azsql.py
@@ -1,0 +1,242 @@
+"""Microsoft Azure SQL Database plugin to read Azure SQL DB data.
+
+This module defines the :class:`AzSQL` class that retrieves SQL DB
+from Microsoft Azure. This module also retrieves the Transparent
+Data Encryption (TDE) configuration of the SQL database.
+"""
+
+
+import logging
+
+from azure.common.credentials import ServicePrincipalCredentials
+from azure.mgmt.resource import SubscriptionClient
+from azure.mgmt.sql import SqlManagementClient
+from msrestazure import tools
+
+from cloudmarker import ioworkers, util
+
+_log = logging.getLogger(__name__)
+
+
+class AzSQL:
+    """Azure SQL Database plugin."""
+
+    def __init__(self, tenant, client, secret, processes=4, threads=30,
+                 _max_subs=0, _max_recs=0):
+        """Create an instance of :class:`AzSQL` plugin.
+
+         Note: The ``_max_subs`` and ``_max_recs`` arguments should be
+         used only in the development-test-debug phase. They should not
+         be used in production environment. This is why we use the
+         convention of beginning their names with underscore.
+
+        Arguments:
+            tenant (str): Azure subscription tenant ID.
+            client (str): Azure service principal application ID.
+            secret (str): Azure service principal password.
+            _max_subs (int): Maximum number of subscriptions to fetch
+                data for if the value is greater than 0.
+            _max_recs (int): Maximum number of SQL server records
+                to fetch for each subscription.
+        """
+        self._credentials = ServicePrincipalCredentials(
+            tenant=tenant,
+            client_id=client,
+            secret=secret,
+        )
+        self._tenant = tenant
+        self._processes = processes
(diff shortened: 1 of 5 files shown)


cloudmarker/clouds/azsql.py
```diff
@@ -0,0 +1,231 @@
...
+        db_list = sql_client.databases.list_by_server(rg_name, server_name)
+        for db_index, db in enumerate(db_list):
+            sql_db_name = db.as_dict().get('name')
+            tde_config = \
+                sql_client.transparent_data_encryptions.get(rg_name,
+                                                            server_name,
+                                                            sql_db_name)
+            _log.info('Found sql_db #%d: %s: %s; %s',
+                      db_index, sql_db_name, server_name,
+                      sub_summ)
```
susam: I suggest writing this log in this format: `'Found sql_db #%d: %s; SQL server #%d: %s; %s'`.

This would make it clear while reading the logs which part is the DB name and which part is the server name.

cloudmarker/clouds/azsql.py
```diff
@@ -0,0 +1,231 @@
...
+                    tools.parse_resource_id(sql_server_id)['resource_group']
+                yield (sub, rg_name, sql_server_name,
+                       util.outline_az_sub(sub_index, sub, tenant))
+
+                # Break after pulling data for self._max_recs number
+                # of SQL servers for a subscriber. Note that if
+                # self._max_recs is 0 or less, then the following
+                # condition never evaluates to True.
+                if server_index + 1 == self._max_recs:
+                    _log.info('Stopping get_subscription_dbs fetch due '
```
susam: Please rewrite this to `Stopping SQL server fetch due ` to be consistent with `AzVM` as well as with the `Found` and `Failed to fetch` logs in this plugin.

cloudmarker/clouds/azsql.py
```diff
@@ -0,0 +1,231 @@
...
+
+        Yields:
+            dict: An Azure SQL server record with TDE configuration.
+
+        """
+        sub_id = sub.get('subscription_id')
+        creds = self._credentials
+        sql_client = SqlManagementClient(creds, sub_id)
+        db_list = sql_client.databases.list_by_server(rg_name, server_name)
+        for db_index, db in enumerate(db_list):
```
susam: Set `db = db.as_dict()` as the first statement. Then we won't have to worry about performing `.as_dict()` everywhere in the code below.

cloudmarker/clouds/azsql.py
```diff
@@ -0,0 +1,231 @@
...
+            sql_db_name = db.as_dict().get('name')
+            tde_config = \
+                sql_client.transparent_data_encryptions.get(rg_name,
+                                                            server_name,
+                                                            sql_db_name)
```
susam: As soon as we get a response, let us change it to `dict` with the `.as_dict()` call so that we don't have to worry about a variable containing native GCP SDK objects anywhere else in the code.

cloudmarker/clouds/azsql.py
```diff
@@ -0,0 +1,231 @@
...
+        """
+        try:
+            tenant = self._tenant
+            creds = self._credentials
+            sub_id = sub.get('subscription_id')
+            sql_client = SqlManagementClient(creds, sub_id)
+            db_server_list = sql_client.servers.list()
+
+            for server_index, sql_server in enumerate(db_server_list):
+                sql_server = sql_server.as_dict()
```
susam: This `sql_server` response that was received is not being yielded by this plugin so that it could be saved in the stores. I think this is useful information we are discarding. Do you have any thoughts around saving this information? It does not have to be in this pull request but I just want to know your thoughts regarding this.

mitprasoon: I gave it some though and have plan to add it to `AzCloud` plugin. `AzCloud` retrieves data about `MySQL` and `PostgreSQL` servers. I plan to add `SQL` sever there for consistency.

susam: Yes, that sounds good.

PR #152, opened 2019-07-01
Add AzSQLDatabaseTDEEvent plugin

mitprasoon:
The `AzSQLDatabaseTDEEvent` evaluates records of type 'sql_db'
for Azure cloud and generates event of type `sql_db_tde_event`
if Transparent Data Encryption (TDE) is disabled on the SQL database
being evaluated.

TDE protects data `at-rest` by encrypting data and logs using AES and
3DES encryption algorithms without changing existing applications.

### DIFF
--- a/cloudmarker/baseconfig.py
+++ b/cloudmarker/baseconfig.py
@@ -41,6 +41,9 @@ plugins:
   rdbmsenforcetlsevent:
     plugin: cloudmarker.events.rdbmsenforcetlsevent.RDBMSEnforceTLSEvent
 
+  azsqldatabasetdeevent:
+    plugin: cloudmarker.events.azsqldatabasetdeevent.AzSQLDatabaseTDEEvent
+
   mockevent:
     plugin: cloudmarker.events.mockevent.MockEvent
 
--- a/cloudmarker/test/test_azsqldatabasetdeevent.py
+++ b/cloudmarker/test/test_azsqldatabasetdeevent.py
@@ -1,0 +1,71 @@
+"""Tests for AzSQLDatabaseTDEEvent plugin."""
+
+
+import copy
+import unittest
+
+from cloudmarker.events import azsqldatabasetdeevent
+
+base_record = {
+    'com':  {
+        'cloud_type': 'azure'
+    },
+    'ext': {
+        'record_type': 'sql_db',
+        'tde_enabled': True
+    }
+}
+
+
+class AzSQLDatabaseDisabledTDEEventTest(unittest.TestCase):
+    """Tests for AzSQLDatabaseTDEEvent plugin."""
+
+    def test_com_bucket_missing(self):
+        record = copy.deepcopy(base_record)
+        record['com'] = None
+        plugin = azsqldatabasetdeevent.AzSQLDatabaseTDEEvent()
+        events = list(plugin.eval(record))
+        self.assertEqual(events, [])
+
+    def test_cloud_non_azure(self):
+        record = copy.deepcopy(base_record)
+        record['com']['cloud_type'] = 'non_azure'
+        plugin = azsqldatabasetdeevent.AzSQLDatabaseTDEEvent()
+        events = list(plugin.eval(record))
+        self.assertEqual(events, [])
+
+    def test_ext_bucket_missing(self):
+        record = copy.deepcopy(base_record)
+        record['ext'] = None
+        plugin = azsqldatabasetdeevent.AzSQLDatabaseTDEEvent()
+        events = list(plugin.eval(record))
+        self.assertEqual(events, [])
+
+    def test_record_type_non_sql_db(self):
+        record = copy.deepcopy(base_record)
+        record['ext']['record_type'] = 'non_sql_db'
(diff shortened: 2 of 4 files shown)


cloudmarker/events/azsqldatabasedisabledtdeevent.py
```diff
@@ -0,0 +1,95 @@
...
+
+import logging
+
+from cloudmarker import util
+
+_log = logging.getLogger(__name__)
+
+
+class AzSQLDatabaseDisabledTDEEvent:
+    """Az SQL Database TDE event plugin."""
```
susam: Let us start spelling out `Az` in full as `Azure` because this is going to be part of publicly visible documentation.

If you would like to make the same change in other plugins, that would be acceptable as part of this pull request but it is not necessary.

I think the `Database` here should also be in lowercase like `database` because we are using it just like any other common word (not really a brand name).

cloudmarker/events/azsqldatabasedisabledtdeevent.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+import logging
+
+from cloudmarker import util
+
+_log = logging.getLogger(__name__)
+
+
+class AzSQLDatabaseDisabledTDEEvent:
```
susam: I think a more handy name would be `AzSQLDatabaseTDEEvent`. Our plugin names have been devoid of the status so far ("disabled" or "enabled"), so we can continue that style by removing `Disabled` from the name. Note: An update to this would require an update in the module filename too.

cloudmarker/events/azsqldatabasedisabledtdeevent.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def eval(self, record):
+        """Evaluate Azure SQL DB for disabled TDE.
+
+        Arguments:
+            record (dict): A SQL DB record.
+
+        Yields:
+            dict: An event record representing an Azure SQL DB with
+            TDE disabled
```
susam: Please end this with full stop for consistency.

cloudmarker/events/azsqldatabasedisabledtdeevent.py
susam: This plugin is a good candidate for having a built-in config key in `baseconfig.py`. Could you please add it so that it can be referred to easily in user's configuration?

mitprasoon: I have added `AzSQLDatabaseTDEEvent` to `baseconfig.py` as a built-in config key.

cloudmarker/events/azsqldatabasedisabledtdeevent.py
mitprasoon: Fixed

cloudmarker/events/azsqldatabasedisabledtdeevent.py
mitprasoon: Renamed to `AzSQLDatabaseTDEEvent `

cloudmarker/events/azsqldatabasedisabledtdeevent.py
mitprasoon: Addressed

cloudmarker/test/test_azsqldatabasetdeevent.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+import copy
+import unittest
+
+from cloudmarker.events import azsqldatabasetdeevent
+
+base_record = {
+    'com':  {
+        'record_type': 'database',
```
susam: The `com` `record_type` does not appear to be necessary for this test, so I suggest removing this field from the test record.

cloudmarker/baseconfig.py
```diff
@@ -41,6 +41,9 @@
   rdbmsenforcetlsevent:
     plugin: cloudmarker.events.rdbmsenforcetlsevent.RDBMSEnforceTLSEvent
 
+  AzSQLDatabaseTDEEvent:
+    plugin: cloudmarker.events.azsqldatabasetdeevent.AzSQLDatabaseTDEEvent
```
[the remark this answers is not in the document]
mitprasoon: Corrected to `azsqldatabasetdeevent `

pylama.ini
```diff
@@ -116,4 +116,7 @@ ignore = R0913,W0703,R0201
 # W0703 Catching too general exception Exception [pylint]
 # R0201 Method could be a function [pylint]
 
+[pylama:cloudmarker/events/azsqldatabasetdeevent.py]
+ignore = R0201
 
+# R0201 Method could be a function [pylint]
```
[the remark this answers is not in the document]
mitprasoon: Added a new line to at the end of the file

cloudmarker/test/test_azsqldatabasetdeevent.py
mitprasoon: Removed `record_type` from the `com` bucket as it was not used. Thanks for pointing this.

PR #154, opened 2019-07-05
Add AzWebApp and AzWebAppTLSEvent plugins

susam:
Author of this commit: @renusrijith

### DIFF
--- a/cloudmarker/events/azwebapptlsevent.py
+++ b/cloudmarker/events/azwebapptlsevent.py
@@ -1,0 +1,95 @@
+"""Microsoft web app minimum TLS version event.
+
+This module defines the :class:`AzWebAppTLSEvent` class that identifies
+a web app with minimum TLS version not equal to the latest version
+(1.2). This plugin works on the web apps config properties found in the
+``com`` bucket of ``web_app`` records.
+"""
+
+
+import logging
+
+from cloudmarker import util
+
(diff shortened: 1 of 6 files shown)


cloudmarker/events/azwebapptlsevent.py
```diff
@@ -0,0 +1,95 @@
...
+        ext = record.get('ext')
+        if ext is None:
+            return
+
+        if ext.get('record_type') != 'web_app_config':
+            return
+
+        min_tls_version = ext.get('min_tls_version')
+
+        if min_tls_version != '1.2':
```
mitprasoon: The key `min_tls_version` is a string. So this expression will be true only of the value of the key `min_tls_version` is 1.2
This will evaluate to false even if `min_tls_version` is 1.3 
TLS 1.3 is already made available to its customers by many cloud vendors. I suggest to make this check configurable with 1.2 set to minimum tls version for now.

susam: @mitprasoon Thank you for highlighting this issue. Since this code has come as a contribution from @renusrijith who is not currently a member of this GitHub org, I will go ahead and merge this commit as is. I don't want to modify her commit because then the notion of the author of this commit would become ambiguous. However, this is an importnat issue we have to address, so I have created issue #160 to track this.

PR #157, opened 2019-07-08
Modify docs for GCPCloud plugin

jaibhageria:


docs/tutorial.rst
```diff
@@ -605,6 +595,9 @@ Follow these steps to get started with auditing a GCP cloud environment.
      plugin referred to with the key name ``firewallruleevent`` in the
      `built-in base config`_.
 
+Note that the ``GCPCloud`` plugin audits all the projects under the
+``service-account`` for the ``keyfile.json`` provided and also discovers
```
susam: The word "service-account" is not a code fragment or file name or anything that exists literally on computer. It's a normal English word, so this should not use the inline code syntax.

PR #166, opened 2019-07-24
Add AzLogProfileMissingCategoryEvent plugin

mitprasoon:
A log profile controls how the activity log is exported. Configure the
log profile to collect logs for the categories `write`, `delete` and
`action` ensures that all the control/management plane activities
performed on the subscription are exported.

The `AzLogProfileMissingCategoryEvent` plugin evaluates records of
type `log_profile` for Azure cloud and generates event of type
`log_profile_missing_category_event` if a log profile is not
enabled for all three activity types i.e. `Write`, `Delete` and
`Action`.

### DIFF
--- a/cloudmarker/events/azlogprofilemissingcategoryevent.py
+++ b/cloudmarker/events/azlogprofilemissingcategoryevent.py
@@ -1,0 +1,123 @@
+"""Microsoft Azure Log Profile Missing Category Type Event.
+
+This module defines the :class:`AzLogProfileMissingCategoryEvent` class
+that identifies if a log profile which is not enable for all the categories
+i.e. Write, Delete and Action. This plugin works on the log profile
+properties found in the ``raw`` bucket of ``log_profile`` records.
+"""
+
+
+import logging
+
+from cloudmarker import util
+
+_log = logging.getLogger(__name__)
+
+
+class AzLogProfileMissingCategoryEvent:
+    """Azure log profile missing category event plugin."""
+
+    def __init__(self):
+        """Create an instance of the class.
+
+        Create an instance of the
+        :class:`AzLogProfileMissingCategoryEvent`.
+        """
+
+    def eval(self, record):
+        """Evaluate Azure log profiles for enabled categories.
+
+        Arguments:
+            record (dict): An Azure log profile record.
+
+        Yields:
+            dict: An event record representing an Azure log profile
+            which is not enabled for all categories.
+
+        """
+        com = record.get('com', {})
+        if com is None:
+            return
+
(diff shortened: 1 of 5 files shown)


cloudmarker/events/azlogprofilemissingactivitytypeevent.py
```diff
@@ -0,0 +1,120 @@
...
+        raw (dict): Log profile record `raw` bucket
+    Yields:
+        dict: An event record representing a log profile which is not enabled
+              for an activity.
+
+    """
+    activities = ('Write', 'Delete', 'Action')
+    for activity in activities:
+        if activity not in raw.get('categories'):
+            yield _get_log_profile_activity_event(com, ext, activity)
```
susam: We both discussed that it would be better to generate one event for one log profile as opposed to generating one event for each activity type.

cloudmarker/events/azlogprofilemissingactivitytypeevent.py
```diff
@@ -0,0 +1,120 @@
...
+        missing_activity (string): Missing activity type
+    Returns:
+        dict: An event record representing SQL DB with disabled TDE
+
+    """
+    friendly_cloud_type = util.friendly_string(com.get('cloud_type'))
+    reference = com.get('reference')
+    description = (
+        '{} log profile {} has is not enabled for {} activity.'
+        .format(friendly_cloud_type, reference, missing_activity)
```
susam: I think a good description would be:

```python
'{} log profile {} does not include categories {}.'.format(..., ..., util.friendly_list(missing_activities))
```

Similarly for recommendation:

```
'Check {} log profile {} and enable categories {}.'.format(..., ..., util.friendly_list(missing_activities))
```

cloudmarker/events/azlogprofilemissingactivitytypeevent.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+import logging
+
+from cloudmarker import util
+
+_log = logging.getLogger(__name__)
+
+
+class AzLogProfileMissingActivityTypeEvent:
```
susam: I would like this class name to follow the Azure terminology of "category". The key name in the JSON response is `categories`. The CIS benchmark document for Azure also refers to it as "category". My suggestion: `AzLogProfileMissingCategoryEvent`.

The record type would then become: `log_profile_missing_category_event`. I think these names are a little bit neater.

mitprasoon: Renamed to `AzLogProfileMissingCategoryEvent ` also event/record type changed to `log_profile_missing_category_event `

PR #167, opened 2019-07-26
Add AzLogProfileRetentionEvent plugin

mitprasoon:
The `AzLogProfileRetentionEvent` plugin evaluate an Azure `log_profile`
record and generate and event of type `log_profile_retention_event` if
the log profile's retention policy is configured for less number of days
than required. The default value for minumum number of days is 365, as
per CIS recommendation but can be changed to any other desired value
through the `min_retention_days` parameter.

It is to be noted that setting retention (days) to 0 from portal
sets days to 0 and enabled to false, which will retain the data forever.

### DIFF
--- a/cloudmarker/events/azlogprofileretentionevent.py
+++ b/cloudmarker/events/azlogprofileretentionevent.py
@@ -1,0 +1,113 @@
+"""Microsoft Azure Log Profile Retention Event.
+
+This module defines the :class:`AzLogProfileRetentionEvent` class that
+identifies if an Azure log profile's retention policy is configured for
+less than the minimum number of days than required. This plugin works
+properties found in the ``ext`` bucket of ``log_profile`` records.
+"""
+
+
+import logging
+
+from cloudmarker import util
+
+_log = logging.getLogger(__name__)
+
+
+class AzLogProfileRetentionEvent:
+    """Azure log profile retention event plugin."""
+
+    def __init__(self, _min_retention_days=365):
+        """Create an instance of :class:`AzLogProfileRetentionEvent`.
+
+        Arguments:
+            _min_retention_days (int): Minimum required
+                                    retention days.
+
+        """
+        self._min_retention_days = _min_retention_days
+        _log.info("Initialized; minimum retention days: %d",
+                  self._min_retention_days)
+
+    def eval(self, record):
+        """Evaluate Azure log profiles for retention policy.
+
+        Arguments:
+            record (dict): An Azure log profile record.
+
+        Yields:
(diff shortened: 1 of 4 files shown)


mitprasoon:
Added `azlogprofileretentionevent ` to `baseconfig`

cloudmarker/events/azlogprofileretentionevent.py
```diff
@@ -0,0 +1,116 @@
...
+
+        ext = record.get('ext', {})
+        if ext is None:
+            return
+
+        if ext['retention_enabled']:
+            if ext['retention_days'] <\
+                    self._min_retention_days:
+                yield from \
+                    _get_log_profile_retention_event(com, ext,
```
susam: These statements should be formatted like this:

```python
        if ext['retention_enabled']:
            if ext['retention_days'] < self._min_retention_days:
                yield from _get_log_profile_retention_event(
                    com, ext, self._min_retention_days)
        else:
            if ext['retention_days'] == 0:
                yield from _get_log_profile_retention_event(
                    com, ext, self._min_retention_days)
```

The use of `\` for line-continuation is discouraged in PEP-0008. It's use is reserved for cases where there is no other alternative possible, e.g., a long `with` statement.

cloudmarker/events/azlogprofileretentionevent.py
```diff
@@ -0,0 +1,116 @@
...
+                                                     self._min_retention_days)
+        else:
+            if ext['retention_days'] == 0:
+                yield from \
+                    _get_log_profile_retention_event(com, ext,
+                                                     self._min_retention_days)
```
susam: I am confused about the `else` part. If we have reached the `else` part, it means that `ext['retention_enabled']` is `False`. In that case, should we not generate an event regardless of what the actual number of retention days is?

cloudmarker/events/azlogprofileretentionevent.py
mitprasoon: @susam, I have rectified this.

cloudmarker/events/azlogprofileretentionevent.py
mitprasoon: To store logs for forever `retention_enabled` must be `False` and `retention_days` must be 0.
The `else` here takes care of that scenario, if the `retention_enabled` is `False` and `retention_days` is not `0` then an event is generated.

susam:
This looks good. Could you please update your local `master` with the latest `master` of this repo, then rebase your pull request on the latest `master`, and force-push your branch?

PR #187, opened 2020-03-02
Add AzStorageAccountDefaultNetworkAccessEvent plugin

mitprasoon:
Added AzStorageAccountDefaultNetworkAccessEvent event plugin. This plugin evaluates an Azure storage acccount and generated an event of type `storage_account_default_network_access_event` if the storage account has default network access set to `Allow`.

Storage account access should be limited to a selected networks. To do so the default network access must be set to `Deny`.

### DIFF
--- a/pylama.ini
+++ b/pylama.ini
@@ -226,3 +226,9 @@ ignore = R0201
 ignore = R0201
 
 # R0201 Method could be a function [pylint]
+
+[pylama:cloudmarker/events/azstorageaccountdefaultnetworkaccessevent.py]
+ignore = R0201
+
+# R0201 Method could be a function [pylint]
+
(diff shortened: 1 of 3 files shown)


pylama.ini
```diff
@@ -226,3 +226,8 @@ ignore = R0201
 ignore = R0201
 
 # R0201 Method could be a function [pylint]
+
+[pylama:cloudmarker/events/azstorageaccountdefaultnetworkaccessevent.py]
+ignore = R0201
+
+# R0201 Method could be a function [pylint]
```
susam: Please add a newline at the end of this file.

mitprasoon: Addressed.

PR #188, opened 2020-03-03
Add AzStorageAccountAllowTrustedServicesEvent

mitprasoon:
Added `AzStorageAccountAllowTrustedServicesEvent` event plugin. This
plugin evaluates Azure storage accounts and generates an event if
Microsoft Azure services are not allowed access.

From `CIS Microsoft Azure Foundations Benchmark v1.1.0`:

Some Microsoft services that interact with storage accounts operate
from networks that can't be granted access through network rules. To
help this type of service work as intended, allow the set of trusted
Microsoft services to bypass the network rules. These services will
then use strong authentication to access the storage account. If the
Allow trusted Microsoft services exception is enabled, the following
services: Azure Backup, Azure Site Recovery, Azure DevTest Labs,
Azure Event Grid, Azure Event Hubs, Azure Networking, Azure Monitor
and Azure SQL Data Warehouse (when registered in the subscription),
are granted access to the storage account.

### DIFF
--- a/cloudmarker/clouds/azstorageaccount.py
+++ b/cloudmarker/clouds/azstorageaccount.py
@@ -212,6 +212,13 @@ def _process_storage_account_properties(storage_account_index,
     default_network_access_allowed = True
     if storage_account['network_rule_set'].get('default_action') != 'Allow':
         default_network_access_allowed = False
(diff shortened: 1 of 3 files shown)


mitprasoon:
Thanks @susam

PR #190, opened 2020-03-05
Azure VM extension plugin

mitprasoon:
Extended the `AzVM` cloud plugin to extract the extenstion names from the VM instance view details and put it in the `ext` bucket. This extension information can be used to audit (whitelisted, blacklisted,
required) extensions installed on the Azure VMs.
Added `AzVMExtensionEvent` plugin to evaluate and audit added Azure VM extensions. This plugin takes list of whitelisted, blacklisted and required VM extensions as configuration parameter and generate events if any of those conditions are not met.

### DIFF
--- a/cloudmarker/events/azvmextensionevent.py
+++ b/cloudmarker/events/azvmextensionevent.py
@@ -1,0 +1,224 @@
+"""Microsoft Azure VM extension event.
+
+This module defines the :class:`AzVMExtensionEvent` class that
+evaluates Azure VM extensions. This plugin works on the virtual
+machine properties found in the ``ext`` bucket of ``vm_instance_view``
+records.
+"""
+
+
+import logging
+
+from cloudmarker import util
+
+_log = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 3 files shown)


cloudmarker/events/azvmextensionevent.py
```diff
@@ -0,0 +1,224 @@
...
+class AzVMExtensionEvent:
+    """Az VM Data extension event plugin."""
+
+    def __init__(self, _whitelisted=None, _blacklisted=None, _required=None):
+        """Create an instance of :class:`AzVMExtensionEvent`.
+
+        Arguments:
+            _whitelisted (list): List of whitelisted extensions.
+            _blacklisted (list): List of blacklisted extensions.
+            _required (list): List of required extensions.
```
susam: These can be non-underscore variables, i.e., users can override these variables in their configuration if they wish to.

mitprasoon: Thanks for your review suggestion @susam . I have addressed this.

